PR #11, opened 2015-12-02
Price multiplier was not being set correctly in mkEC2UserDataSlave.

dinatale2:
Price multiplier was not being set correctly in mkEC2UserDataSlave. Now correctly assigning price multiplier if it is provided by the caller.


behlendorf:
@dinatale2 thanks!  I've applied it.  For future reference when opening a pull request it's generally best practice to make your change on a branch and not master.

PR #15, opened 2015-12-17
Set overlay=on on zfsstress datasets

nedbass:
The scripts don't always coordinate properly when zfsstress exports and
reimports the test pool. Some files may then get created at the
mountpoint for the test dataset, and this prevents it from getting
remounted unless overlay=on is set.

Signed-off-by: Ned Bass <bass6@llnl.gov>


behlendorf:
LGTM

nedbass:
Tacked on another commit to this PR, as it also relates to letting the test resume correctly after the pool is exported.

behlendorf:
They both LGTM

PR #16, opened 2016-01-04
boot into debug kernel when BB_KERNEL_TYPE == DEBUG

ofaaland:
Alter bootstrap script to install debug kernel and boot into it prior to
beginning build and test, on Redhat-based distros, when
BB_KERNEL_TYPE is set to DEBUG.

This will be paired with a new test step that checks /proc or dmesg for
lockdep warnings produced while testing zfs.

### DIFF
--- a/scripts/bb-bootstrap.sh
+++ b/scripts/bb-bootstrap.sh
@@ -48,6 +48,9 @@ fi
 if test ! "$BB_USE_PIP"; then
     BB_USE_PIP=0
 fi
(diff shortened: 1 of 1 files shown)


behlendorf:
Let's go with this for now and get this infrastructure in place.  It can always be optimized/reworked if needed.  Merged as:

fe98de5 boot into debug kernel when BB_KERNEL_TYPE == DEBUG

PR #19, opened 2016-02-04
Optimize how slaves are assigned builds.

dinatale2:
It is preferable to assign a build to a slave who is
idle (already spun up and waiting) over a slave that has yet
to be started or a busy one. If there are no idle slaves,
and there is one that can be spun up, then spin one up.
Otherwise, randomly choose among our busy slaves.

### DIFF
--- a/master/builderinfo.py
+++ b/master/builderinfo.py
@@ -209,10 +209,31 @@ class BuilderInfo(object):
     def makeSlaves(self):
         return [info.makeBuildSlave() for info in self.slaves]
 
+    @staticmethod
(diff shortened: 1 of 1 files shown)


behlendorf:
Nice, I see this change depends on `isIdle()` which is fine but be aware it's a function we've added to our slightly patched version of buildbot.  This LGTM I'll see about rolling it out when the buildbot's next idle.

behlendorf:
Merged as:

ea3ffad Optimize how slaves are assigned builds.

PR #20, opened 2016-02-12
Force root volumes to be deleted upon termination on EC2 volumes.

dinatale2:
This patch is for buildbot. It introduces the ability to set all root
volumes of an instance to be deleted upon termination. This change
does not impact additionally attached volumes.


behlendorf:
Looks good and I didn't see any issues while testing.  Squashed and merged as:

7922e27 Force root volumes to be deleted upon termination on EC2 volumes.

PR #21, opened 2016-02-24
Improved how builders are prioritized when assigned changes.

dinatale2:
Rewrote the prioritizeBuilders function in master.cfg.
Builders that have an idle slave should be chosen before all
others. If no idle builder exists, then choose a builder
which has no substantiated slaves. Substantiation takes on the
order of 3-4 minutes. During that time, another builder's slave
may become idle. Finally, just assign to a busy builder since
all builders must be busy.


behlendorf:
This looks like a nice improvement, it will be interesting to see how it works out in practice with a real workload.  It's been merged and is running on the buildbot master now.

80d75df Improved how builders are prioritized when assigned changes.

PR #22, opened 2016-02-25
Addition of SUSE Linux spot instances.

dinatale2:
Buildbot uses the request pricing history feature
to determine the average price for Linux/Unix type
instances only (it's hardcoded). Introduced a way
to specify the type of instance you wish to create
to the EC2 Latent Build Slave object. Also, added
a Large Slave Info object in builderinfo.py.

Be sure to apply the patch to buildbot before using
the new SUSE Large Slave Info object.

### DIFF
--- a/master/builderinfo.py
+++ b/master/builderinfo.py
@@ -76,6 +76,7 @@ class EC2SlaveInfo(BuildSlaveInfo):
         region="us-west-2"
         placement=None
         spot_instance=False
+        product_description="Linux/UNIX (Amazon VPC)"
         max_spot_price=0.08
         price_multiplier=1.25
         keepalive_interval=3600
@@ -131,6 +132,8 @@ runurl """ + self.url + """bb-bootstrap.sh
                 price_multiplier=kwargs[key]
             if key=="placement":
                 placement=kwargs[key]
+            if key=="product_description":
+                product_description=kwargs[key]
 
         return EC2LatentBuildSlave(
             self.name, self.password, instance_type, ami=self.ami,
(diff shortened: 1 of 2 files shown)


dinatale2:
I would like to mention that my patch to buildbot's EC2 latent slave is a bit "hacky". The proper way to implement it likely would be to refactor buildbot's EC2 support into an abstract ec2 latent slave, then derive from there each type of product you wish to support. That would be more time consuming to achieve. I may try and tackle that one later.

behlendorf:
Or better yet if the product description be queried from the AMI that should be used by default.  That would remove the need for the caller to be aware of this internal detail when configuring the latent slave.  You could additionally allow for this value to be passed in but it should not be required to get reasonable default behavior.

dinatale2:
Linux/UNIX (Amazon VPC) is set by default in the EC2 latent slave constructor. I didn't see a way to use the AMI to get the product description when searching EC2 documentation, unfortunately. I would assume that means the EC2 API doesn't have a way to retrieve that information.

behlendorf:
This looks like a thing you can do through the command line utilities.  As for the library I'm not sure.  It would be worth doing a little more investigation before determine this can't be done.

http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/paid-amis.html#using-paid-amis-finding-paid-ami

dinatale2:
Are you talking about the product code? Because I don't think the product code is the same as the product description.

behlendorf:
It looks like you're right.  According to [this thread](https://forums.aws.amazon.com/thread.jspa?threadID=150594) on the developer forum about there isn't a way to get the product description except through the web interface.  Bummer, then it does seem like we need to pass it in.  I withdraw my objection, but I'm a bit disappointed in ec2.

dinatale2:
Would it be worth closing this pull request and opening an issue? This branch is very stale and the only thing you would really need is the patch from it since you are aren't using SUSE builders right now. Unless you are starting to use the OpenSFS buildbot fork, in which case, you already have this change.

behlendorf:
@dinatale2 good idea.  Let's close this out as stale, we can address this issue when/if we setup an OpenSUSE builder.

PR #23, opened 2016-04-19
Refactored how slaves and builder are created.

dinatale2:
Instead of having a slave info and builder info class,
directly derive our new classes from existing buildbot
classes. Works out to be much cleaner.


dinatale2:
This should at least pass a checkconfig. I can't fully test this. If you see or find any issues, please let me know and I will correct them.

behlendorf:
Nice work!  Any patch the preserves all the existing functionality and is a net negative 300 lines of code is a definitely a step is the right direction.  I've applied this to the buildbot and everything seems to be working properly, I'll keep an eye on it for a while just to make sure.  Thanks!

master/master.cfg
```diff
@@ -349,479 +350,311 @@ def prioritizeBuilders(buildmaster, builders):
...
-                "Debian-8-arm-buildslave1",
-                slave_userpass["Debian-8-arm-buildslave1"]),
-            PersistSlaveInfo(
-                "Debian-8-arm-buildslave2",
-                slave_userpass["Debian-8-arm-buildslave2"]),
-        ],
+    ZFSBuilderConfig(
+        name="Debian 8 arm (BUILD)",
+        factory=build_factory,
+        slavenames=[slave.name for slave in debian8_arm_slave],
```
behlendorf: This should be `slave.slavename`, I've made the fix locally and can squash it in to this patch.  No need to refresh this.  This was the only issues I've found thus far.

dinatale2: I wonder if I can remove slave.name and replace them all with slave.slavename since they are all derived from a buildbot slave class now...

behlendorf: That seems to pass checkconfig but I'll need to wait for the current builds to finish before restarting the build master.  I agree, consistency would be good for this.

PR #25, opened 2016-05-12
Bootable image: build bootable image on Fedora.

Rudd-O:
Later on, we can execute the bootable image and test that it, in fact, boots correctly.

### DIFF
--- a/master/master.cfg
+++ b/master/master.cfg
@@ -490,6 +490,13 @@ ubuntu14_x86_64_testslave = [
     ) for i in range(0, numtestslaves)
 ]
 
+fedora23_x86_64_testslave = [
+    ZFSEC2TestSlave(
+        name="Fedora-23-x86_64-testslave%s" % (str(i+1)),
+        ami="ami-f8709e98"
+    ) for i in range(0, numslaves)
+]
+
 #ubuntu14_i686_testslave = [
 #    ZFSEC2PVTestSlave(
 #        name="Ubuntu-14.04-i686-testslave%s" % (str(i+1)),
@@ -649,6 +656,13 @@ test_builders = [
         name="Ubuntu 14.04 x86_64 (TEST)",
         factory=test_factory,
         slavenames=[slave.name for slave in ubuntu14_x86_64_testslave],
+        tags=test_tags,
+        properties=builder_default_properties,
+    ),
+    ZFSBuilderConfig(
+        name="Fedora 23 x86_64 (TEST)",
+        factory=test_factory,
+        slavenames=[slave.name for slave in fedora23_x86_64_testslave],
         tags=test_tags,
         properties=builder_default_properties,
     ),
--- a/master/master.cfg
+++ b/master/master.cfg
@@ -17,7 +17,6 @@ web_url = "http://build.zfsonlinux.org/"
 spl_repo = "https://github.com/zfsonlinux/spl.git"
 zfs_repo = "https://github.com/zfsonlinux/zfs.git"
 linux_repo = "https://github.com/torvalds/linux.git"
-zfs_fedora_installer_repo = "https://github.com/Rudd-O/zfs-fedora-installer.git"
 zfs_path = "/usr/libexec/zfs:/usr/share/zfs:/usr/lib/rpm/zfs:/usr/lib/zfs"
 bin_path = "/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
 
@@ -25,11 +24,9 @@ all_repositories = {
     "https://github.com/torvalds/linux" : 'linux',
     "https://github.com/zfsonlinux/spl" : 'spl',
     "https://github.com/zfsonlinux/zfs" : 'zfs',
-    zfs_fedora_installer_repo : 'zfs-fedora-installer',
     "https://github.com/torvalds/linux.git" : 'linux',
     "https://github.com/zfsonlinux/spl.git" : 'spl',
     "https://github.com/zfsonlinux/zfs.git" : 'zfs',
-    zfs_fedora_installer_repo + ".git" : 'zfs-fedora-installer',
 }
 
 c = BuildmasterConfig = {}
@@ -264,11 +261,6 @@ test_factory.addStep(ShellCommand(
     logfiles={"console" : { "filename" : "console.log", "follow" : False }},
     decodeRC={0 : SUCCESS, 1 : FAILURE, 2 : WARNINGS, 3 : SKIPPED },
     description=["zfsstress"], descriptionDone=["zfsstress"]))
-test_factory.addStep(Git(repourl=zfs_fedora_installer_repo,
-    workdir="build/tests/zfs-fedora-installer",
-    mode="full", method="clobber", codebase="zfs-fedora-installer",
-    logEnviron=False, getDescription=True,
-    description=["cloning"], descriptionDone=["cloned"]))
 test_factory.addStep(ShellCommand(
     workdir="build/tests/zfs-fedora-installer",
     env={'PATH' : bin_path + ":" + zfs_path},
--- a/master/master.cfg
+++ b/master/master.cfg
@@ -17,6 +17,7 @@ web_url = "http://build.zfsonlinux.org/"
 spl_repo = "https://github.com/zfsonlinux/spl.git"
 zfs_repo = "https://github.com/zfsonlinux/zfs.git"
 linux_repo = "https://github.com/torvalds/linux.git"
+zfs_fedora_installer_repo = "https://github.com/Rudd-O/zfs-fedora-installer.git"
 zfs_path = "/usr/libexec/zfs:/usr/share/zfs:/usr/lib/rpm/zfs:/usr/lib/zfs"
 bin_path = "/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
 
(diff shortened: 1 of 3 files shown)


Rudd-O:
This is in response to https://github.com/zfsonlinux/zfs/issues/4555 .

scripts/bb-test-bootable-image.sh
```diff
@@ -0,0 +1,65 @@
...
+OUTPUT_LOG="$PWD/output.log"
+
+case "$BB_NAME" in
+Fedora*)
+    true
+    ;;
+
+*)
+    echo "$BB_NAME unknown platform, skipping image test" >>$OUTPUT_LOG 2>&1
+    exit 18
```
behlendorf: It's not obvious but the exact return values [matter to buildbot](http://docs.buildbot.net/latest/developer/results.html).  It controls how the output is formatted in the [waterfall](http://build.zfsonlinux.org/waterfall) and the overall build result.  This should return a 3, briefly:

  0 - success
  1 - warning
  2 - failure
  3 - skipped

behlendorf:
Awesome!  This is definitely along the right lines but let me make some suggestions inline.

master/master.cfg
```diff
@@ -261,6 +264,21 @@ test_factory.addStep(ShellCommand(
     logfiles={"console" : { "filename" : "console.log", "follow" : False }},
     decodeRC={0 : SUCCESS, 1 : FAILURE, 2 : WARNINGS, 3 : SKIPPED },
     description=["zfsstress"], descriptionDone=["zfsstress"]))
+test_factory.addStep(Git(repourl=zfs_fedora_installer_repo,
+    workdir="build/tests/zfs-fedora-installer",
+    mode="full", method="clobber", codebase="zfs-fedora-installer",
+    logEnviron=False, getDescription=True,
+    description=["cloning"], descriptionDone=["cloned"]))
```
behlendorf: This isn't a bad way to go but it's probably more flexible to adopt the approach taken in [bb-test-xfstests.sh](https://github.com/zfsonlinux/zfs-buildbot/blob/master/scripts/bb-test-xfstests.sh) where the script just wget's a tarball from github.  This has a few advantages:

* It can be all done in a single build step and is largely self contained to one script.
* We can point it by default at a specific version or the head of your git repository.
* Setting TEST_FEDORA_INSTALLER_URL in your commit message can be used to override the default.

master/master.cfg
```diff
@@ -261,6 +264,21 @@ test_factory.addStep(ShellCommand(
...
+test_factory.addStep(ShellCommand(
+    workdir="build/tests/zfs-fedora-installer",
+    env={'PATH' : bin_path + ":" + zfs_path},
+    command=["runurl", bb_url + "bb-test-bootable-image.sh"],
+    haltOnFailure=False, maxTime=14400, sigtermTime=30, logEnviron=False,
+    logfiles={"console" : { "filename" : "console.log", "follow" : False },
```
behlendorf: Keeping a console log here probably isn't very important.  It's done for the other testing because it's not unlikely we'll panic the system during testing.  That said, we can certainly keep it.

master/master.cfg
```diff
@@ -261,6 +264,21 @@ test_factory.addStep(ShellCommand(
...
+              "log"     : { "filename" : "output.log",  "follow" : True }},
+    decodeRC={0 : SUCCESS, 1 : FAILURE, 2 : FAILURE, 5 : FAILURE, 9 : FAILURE, 18 : SKIPPED },
```
behlendorf: Wait?  Ignore my comment before, why not just use the defaults?

scripts/bb-test-bootable-image.sh
```diff
@@ -0,0 +1,65 @@
...
+    ;;
+
+# Cleanup the pool and restore any modified system state.  The console log
+# is dumped twice to maximize the odds of preserving debug information.
+cleanup()
+{
+    dmesg >$CONSOLE_LOG
+    sudo -E $ZPOOL destroy -f "$TEST_BOOTABLE_IMAGE_POOL" &>/dev/null
+    sudo -E rm -f /etc/zfs/zpool.cache "$TEST_BOOTABLE_IMAGE_VDEV"
+    dmesg >$CONSOLE_LOG
```
behlendorf: The pool cleanup can be removed since we never create a pool, right?

scripts/bb-test-bootable-image.sh
```diff
@@ -0,0 +1,65 @@
...
+
+TEST_BOOTABLE_IMAGE_POOL=${TEST_BOOTABLE_IMAGE_POOL:-"bootable_image"}
+TEST_BOOTABLE_IMAGE_VDEV=${TEST_BOOTABLE_IMAGE_VDEV:-"/var/tmp/bootable_image.img"}
+TEST_BOOTABLE_IMAGE_OPTIONS=${TEST_BOOTABLE_IMAGE_OPTIONS:-""}
+
+echo -n >$OUTPUT_LOG 2>&1
+
+rm -rf rpms/                            >>$OUTPUT_LOG 2>&1
+mkdir -p rpms/                          >>$OUTPUT_LOG 2>&1
+cp -Rv ../spl/*.rpm ../zfs/*.rpm rpms/  >>$OUTPUT_LOG 2>&1
```
behlendorf: This should work if we update the [bb-build-packages.sh](https://github.com/zfsonlinux/zfs-buildbot/blob/master/scripts/bb-build-packages.sh#L50) script to keep the noarch dkms package it built.

master/master.cfg
```diff
@@ -261,6 +264,21 @@ test_factory.addStep(ShellCommand(
...
+    description=["test-bootable-image"], descriptionDone=["test-bootable-image"]))
+
```
behlendorf: A Fedora builder will need to be added to the `test_builders` below and we'll need to add Fedora the test slaves.  Right now we only compile for Fedora we don't running any of the tests there.  Here are the current test builders.

http://build.zfsonlinux.org/waterfall?tag=Testshttp://build.zfsonlinux.org/waterfall?tag=Tests

master/master.cfg
Rudd-O: I would prefer to keep it for a while, then later we can remove it.

master/master.cfg
Rudd-O: ?

scripts/bb-test-bootable-image.sh
Rudd-O: We do create a pool.  But, now that I think of it, the cleanup process should just bomb if it finds the pool, because that means the cleanup in the installer script failed, so that means a general system failure.  (The cleanup in the installer never fails unless there was a grave error like a kernel OOPS.)

master/master.cfg
behlendorf: This was in regards to the decodeRC values.  It's not a very complicated script so I'd suggest sticking to the default exit values used by the other scripts.

master/master.cfg
Rudd-O: This is coming as a commit.  Soon.

Rudd-O: [user@ring2-projects scripts]$ git push
Counting objects: 6, done.
Delta compression using up to 8 threads.
Compressing objects: 100% (6/6), done.
Writing objects: 100% (6/6), 749 bytes | 0 bytes/s, done.
Total 6 (delta 5), reused 0 (delta 0)
To git@github.com:Rudd-O/zfs-buildbot.git
   472e747..83a9bc3  master -> master

master/master.cfg
Rudd-O: Commit coming along...

Rudd-O: [user@ring2-projects scripts]$ git push
Counting objects: 6, done.
Delta compression using up to 8 threads.
Compressing objects: 100% (6/6), done.
Writing objects: 100% (6/6), 526 bytes | 0 bytes/s, done.
Total 6 (delta 5), reused 0 (delta 0)
To git@github.com:Rudd-O/zfs-buildbot.git
   83a9bc3..3ecf67f  master -> master

master/master.cfg
Rudd-O: [user@ring2-projects scripts]$ git push
Counting objects: 4, done.
Delta compression using up to 8 threads.
Compressing objects: 100% (4/4), done.
Writing objects: 100% (4/4), 451 bytes | 0 bytes/s, done.
Total 4 (delta 3), reused 0 (delta 0)
To git@github.com:Rudd-O/zfs-buildbot.git
   3ecf67f..1b8372d  master -> master

scripts/bb-test-bootable-image.sh
Rudd-O: [user@ring2-projects scripts]$ git push
Counting objects: 4, done.
Delta compression using up to 8 threads.
Compressing objects: 100% (4/4), done.
Writing objects: 100% (4/4), 428 bytes | 0 bytes/s, done.
Total 4 (delta 3), reused 0 (delta 0)
To git@github.com:Rudd-O/zfs-buildbot.git
   1b8372d..166b086  master -> master

master/master.cfg
Rudd-O: Kernel ring buffer log issue addressed.

Rudd-O:
I believe the incoming commits address your comments.

behlendorf:
Replaced with #26 where I've squashed your changes and made a few additional fixes.  However, this still doesn't quite work when I ran through the process manually details in the new PR.

PR #26, opened 2016-05-14
Bootable image: build bootable image on Fedora.

behlendorf:
Currently this only builds the bootable image, we can execute
the bootable image latter and test that it, in fact, boots correctly.

### DIFF
--- a/master/master.cfg
+++ b/master/master.cfg
@@ -261,6 +261,16 @@ test_factory.addStep(ShellCommand(
     logfiles={"console" : { "filename" : "console.log", "follow" : False }},
     decodeRC={0 : SUCCESS, 1 : FAILURE, 2 : WARNINGS, 3 : SKIPPED },
     description=["zfsstress"], descriptionDone=["zfsstress"]))
+test_factory.addStep(ShellCommand(
+    workdir="build/tests/zfs-fedora-installer",
+    env={'PATH' : bin_path + ":" + zfs_path},
+    command=["runurl", bb_url + "bb-test-bootable-image.sh"],
+    haltOnFailure=False, maxTime=14400, sigtermTime=30, logEnviron=False,
+    logfiles={"console" : { "filename" : "console.log", "follow" : False },
+              "log"     : { "filename" : "output.log",  "follow" : True }},
+    decodeRC={0 : SUCCESS, 1 : FAILURE, 2 : FAILURE, 5 : FAILURE, 9 : FAILURE, 3 : SKIPPED },
+    description=["test-bootable-image"], descriptionDone=["test-bootable-image"]))
+
 
 test_factory.addStep(ShellCommand(command=["runurl", bb_url + "bb-cleanup.sh"],
     env={'PATH' : bin_path},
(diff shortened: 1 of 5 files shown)


behlendorf:
@Rudd-O perhaps you can debug this.  When running through the buildbot steps manually I'm running in to an issue with install-fedora-on-zfs being unable to set the root password in the chroot.  

You should be able to reproduce this with a stock Fedora 23.  Here's the test procedure, this basically runs through the commands the buildslave will issue.

* Start by booting a pristine Fedora 23 image and run all the updates.

* Setup an environment like the buildslaves.

```
$ pwd
/home/fedora

$ git clone https://github.com/zfsonlinux/spl.git
$ git clone https://github.com/zfsonlinux/zfs.git
$ git clone https://github.com/zfsonlinux/zfs-buildbot.git

$ cd zfs-buildbot
$ git fetch origin pull/26/head:fedora-installer
$ git checkout fedora-installer
$ cd ..

$ mkdir -p tests/zfs-fedora-installer

$ export PATH="$PATH:/usr/share/zfs/"
$ sudo sh -c 'echo BB_NAME=Fedora-23-x86_64-testslave1 >/etc/buildslave'

# Add  '/usr/share/zfs' to secure_path in '/etc/sudoers'
$ vi /etc/sudoers 
```

* Install dependencies.

```
$ sh -x ./zfs-buildbot/scripts/bb-dependencies.sh
```

*  Build the SPL and ZFS packages then install them.

```
$ cd spl
$ sh -x ../zfs-buildbot/scripts/bb-build-packages.sh
$ cd ../zfs
$ sh -x ../zfs-buildbot/scripts/bb-build-packages.sh
```

* Prepare the `tests`, this shouldn't be needed but just to be thorough.

```
$ cd ../tests
$ sh -x ../zfs-buildbot/scripts/bb-test-prepare.sh 
```

* Run the new zfs-fedora-install test case.

```
$ cd zfs-fedora-installer
$ sh -x ../../zfs-buildbot/scripts/bb-test-bootable-image.sh
<FAILURE>
```

* From `output.log`.

```
DEBUG:root:Popening chroot /mnt/bootable_image passwd --stdin root in cwd '/home/fedora/tests/zfs-fedora-installer'
passwd: Authentication token manipulation error
Changing password for user root.
Unexpected error: Command '[['chroot', '/mnt/bootable_image', 'passwd', '--stdin', 'root']]' returned non-zero exit status 1
```

Rudd-O:
I will debug this very soon.  I will replicate the process locally, see
how it goes.  Promise to give you an answer soon.  Just have to get an
actual kernel corresponding to an actual kernel-devel running on my
virtualization systems, which is not something they are amenable to,
because of how these environments boot.  But I will nail this.

-- 
    Rudd-O
    http://rudd-o.com/

Rudd-O:
So, I have a number of things to report:

1. I have successfully built automation to actually run, and boot, ZFS Fedora images within Qubes OS.  This is trickier than you may think, because I need to (a) boot the system that will perform the build process using PVgrub, which is not the default (b) run the entire process inside a `mock` jail, as QEMU is not installable directly in the VM, even if booted from PVgrub.
2. I could not reproduce the irritating PAM error you got.  I simply could not, no matter how I tried.  Note that, because everything was running within the `mock` jail, that could have had an impact on what happened.  In any case, it works.
3. The build process takes a while to set up the `mock` jail the first time, because the packages need to be downloaded and installed to the jail.
4. I put a _hyooge_ number of bug fixes into zfs-fedora-installer.  Those are now merged into the master branch of my repo for that project.
5. The scripts are there.  You do not care to read about how the Qubes OS stuff works, but what you definitely care about, is the `imagebuild` script, which is what does the magic of building everything within the `mock` jail (note that you also care about the `mockwrapper` script, which is used by the `imagebuild` script to do the magic).  `imagebuild` needs to be adapted to the buildbot environment, but it should be fairly straightforward.

The stuff is here:

https://github.com/Rudd-O/zfs-fedora-installer/tree/master/qubes

Check the README.md out, but focus more on the `imagebuild` and `mockwrapper` parts, as those are the parts that can be used most directly in the buildbot environment.  All of that can be adapted to work fine with multiple Fedora releases.

If you have any questions, I am right here.

PR #28, opened 2016-05-20
Prevent Require-spl dependency from persisting to other PRs

### LINKED ISSUE #27: Fix Requires-spl syntax
After submitting a PR with the `Requires-spl:` keyword all subsequent build requests will use the same SPL commit instead of defaulting back to master.  Restarting the build master resets the SPL dependency and is the current work around the issue.

dinatale2:
When a PR is submitted with the Requires-spl keyword, all
subsequent builds will use that SPL commit instead of
defaulting back to master. Corrected the
CustomSingleBranchScheduler.getCodebaseDict function
to make a deep copy of the codebase dictionary before
modifying it.

This closes #27.

### DIFF
--- a/master/master.cfg
+++ b/master/master.cfg
@@ -787,6 +787,7 @@ from buildbot.schedulers.trysched import Try_Userpass
 from buildbot.schedulers.basic import SingleBranchScheduler
 from buildbot.changes.filter import ChangeFilter
 from buildbot.changes import filter
+import copy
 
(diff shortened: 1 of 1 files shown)


behlendorf:
@dinatale2 I've applied this patch to the buildbot and opened PR zfsonlinux/zfs#4678 to test it.  Luckily we happen to have an open SPL PR which needs testing.

behlendorf:
@dinatale2 that seems to have done the trick.  Thanks!

PR #30, opened 2016-09-27
Implement a STYLE builder

dinatale2:
Place linting and style checking into its own builder.
This will provide developers with faster feedback
regarding style issues and remove the redundant checks
that go on for each BUILD builder.

### DIFF
--- a/master/master.cfg
+++ b/master/master.cfg
@@ -73,6 +73,35 @@ def do_step_build_spl(step):
 def do_step_build_zfs(step):
     return do_step_build(step, 'buildzfs')
 
+style_factory = util.BuildFactory()
+
+style_factory.addStep(Git(repourl=zfs_repo, workdir="build/zfs",
+    mode="full", method="clobber", codebase="zfs",
+    logEnviron=False, getDescription=True,
+    description=["cloning"], descriptionDone=["cloned"]))
+style_factory.addStep(ShellCommand(
+    workdir="build/zfs", env={'PATH' : bin_path,
+        'CONFIG_OPTIONS' : '--with-config=srpm'},
+    command=["runurl", bb_url + "bb-build-zfs.sh"],
+    haltOnFailure=True, logEnviron=False,
+    logfiles={"configure"  : { "filename" : "configure.log", "follow" : True },
+              "config.log" : { "filename" : "config.log",    "follow" : True },
+              "make"       : { "filename" : "make.log",      "follow" : True }},
+    decodeRC={0 : SUCCESS, 1 : FAILURE, 2 : WARNINGS, 3 : SKIPPED },
+    description=["building zfs"], descriptionDone=["built zfs"]))
+style_factory.addStep(ShellCommand(command=["make", "checkstyle"],
+    env={'PATH' : bin_path},
+    workdir="build/zfs", logEnviron=False,
+    haltOnFailure=False, flunkOnWarnings=True,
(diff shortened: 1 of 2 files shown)


master/master.cfg
```diff
@@ -73,6 +73,35 @@ def do_step_build_spl(step):
...
+style_factory.addStep(Git(repourl=zfs_repo, workdir="build/zfs",
+    mode="full", method="clobber", codebase="zfs",
+    logEnviron=False, getDescription=True,
+    description=["cloning"], descriptionDone=["cloned"]))
+style_factory.addStep(ShellCommand(command=["./autogen.sh"],
+    env={'PATH' : bin_path},
+    workdir="build/zfs", logEnviron=False,
+    decodeRC={0 : SUCCESS, 1 : FAILURE, 2 : WARNINGS, 3 : SKIPPED },
+    description=["autogen"], descriptionDone=["autogen"]))
+style_factory.addStep(ShellCommand(command=["./configure"],
```
behlendorf: Without the SPL installed this is going to fail.  But we could speed this up and remove the SPL dependency by running `./configure --with-config=srpm`.  This disables most of the configure checks and allows us to use some make targets like `dist`, `cstyle`, etc.  Anything which doesn't actually requires us to compile the code.

master/master.cfg
```diff
@@ -386,6 +401,16 @@ def prioritizeBuilders(buildmaster, builders):
 
 c['prioritizeBuilders'] = prioritizeBuilders
 
+# STYLE BUILD SLAVES
+numstyleslaves = 1
+
+style_slaves = [
+    ZFSEC2BuildSlave(
+        name="Amazon-2015.09-x86_64-styleslave%s" % (str(i+1)),
+        ami="ami-469edd26"
```
behlendorf: s/ami-469edd26/ami-f1d69891/ - I've just updated the Amazon AMI to resolve build failures due to them updating their kernel.

dinatale2: Updated the AMI.

master/master.cfg
dinatale2: Good catch, fixed!

master/master.cfg
```diff
@@ -73,6 +73,35 @@ def do_step_build_spl(step):
...
+style_factory.addStep(ShellCommand(command=["./autogen.sh"],
+    env={'PATH' : bin_path},
+    workdir="build/zfs", logEnviron=False,
+    decodeRC={0 : SUCCESS, 1 : FAILURE, 2 : WARNINGS, 3 : SKIPPED },
+    description=["autogen"], descriptionDone=["autogen"]))
+style_factory.addStep(ShellCommand(command=["./configure", "--with-config=srpm"],
+    env={'PATH' : bin_path},
+    workdir="build/zfs", logEnviron=False,
+    decodeRC={0 : SUCCESS, 1 : FAILURE, 2 : WARNINGS, 3 : SKIPPED },
+    description=["configuring"], descriptionDone=["configure"]))
```
behlendorf: I'd prefer to use the `bb-build-zfs.sh` script for autogen.sh and configure for consistency with the other builds.  It looks like the script should just work if you set the `CONFIG_OPTIONS="--with-config=srpm"` in the environment of the buildstep.  This will cause `make` to be run but since nothing gets compiled it'll be fast.  Plus we'll get some test coverage on the build system functionality.  You should be able to use the zfs build step from the build factory with minimal modification.

dinatale2: Went ahead and revised the build steps to use `bb-build-zfs.sh`. I will update the README shortly.

PR #31, opened 2016-10-03
Install pep8 on STYLE builders

dinatale2:
Ensure that dependencies are installed on
STYLE builders. This ensures that development
tools to do perform style checking and linting are
installed.


behlendorf:
Merged as 9dc001abd490e33d4cce7b34fbce92b017a45b06

PR #32, opened 2016-10-04
Use flake8 instead of pep8

dinatale2:
Use flake8 to perform python style
checking because pep8 has been deprecated.


behlendorf:
Merged as 3e0bd4fca15bef78d9b8451809852d3b3a0de1dc

PR #34, opened 2016-10-18
Fix logging race condition in bb-test-zfstests.sh

dinatale2:
test-runner.py now creates a simlink to the directory
where output is stored. Have buildbot track the
log via the symlink instead of relying on
bb-test-zfstests.sh to find the output directory.


dinatale2:
This can be accepted once zfs#5293 is accepted.

scripts/bb-test-zfstests.sh
```diff
@@ -52,16 +52,9 @@ $ZFSTESTS $TEST_ZFSTESTS_OPTIONS \
...
     -s $TEST_ZFSTESTS_DISKSIZE \
     -r $TEST_ZFSTESTS_RUNFILE &
-CHILD=$!
-
-sleep 1
-TEST_LOG=$(ls -t /var/tmp/test_results/*/log | head -1)
-rm -f log
-ln -s $TEST_LOG log
-
-wait $CHILD
```
behlendorf: How does this work?  Buildbot still expects to find the log file relative to its current working directory which will be some thing like `/var/lib/buildbot/slaves/zfs/<slave-name>/build/tests/zfstests`.  But the `current` symlink should be created in `/var/tmp/test_results/current`.  We still need to keep that `ln` but link it to the known log path.

```
ln -s /var/tmp/test_results/current/log log
```

master/master.cfg
```diff
@@ -303,7 +303,7 @@ test_factory.addStep(ShellCommand(
     command=["runurl", bb_url + "bb-test-zfstests.sh"],
     haltOnFailure=False, maxTime=14400, sigtermTime=30, logEnviron=False,
     logfiles={"console" : { "filename" : "console.log", "follow" : False },
-              "log"     : { "filename" : "log",         "follow" : True }},
+              "log"     : { "filename" : "current/log",         "follow" : True }},
```
behlendorf: This shouldn't need to change.

scripts/bb-test-zfstests.sh
dinatale2: zfsonlinux/zfs#5293 made it so the current symlink is made in the current working directory. The script no longer has to generate a link, it should be there and named `current`.

behlendorf: Oh I overlooked that when reviewing zfsonlinux/zfs#5293, that's not really what we want.  We want ed the symlink to be created in the `/var/tmp/test_results/` directory next to the dated directories.  This has the advantage of always being writable and predictable.

dinatale2: Ah, my misunderstanding. I'll go fix test-runner real quick, then I can refresh this patch.

dinatale2:
Updated to use the new symlink that test-runner will create in zfsonlinux/zfs#5314

behlendorf:
Merged as:

2f0f996 Enable ShellCommand lazy logging
9efcf8d Hide test steps if they are skipped
5f5add4 Fix logging race condition in bb-test-zfstests.sh

PR #36, opened 2016-11-01
Allow builds to run on specific BUILD builders

dinatale2:
Place the Distribution and Architecture BUILD builders
on separate schedulers so they can be started independently.
Both schedulers will still accept the `build` category.


dinatale2:
Updated the README, missed a spot where all options should be listed.

PR #37, opened 2016-11-02
Introduce a performance builder

dinatale2:
Add a builder to run the performance regression
suite. Use an EC2 d2.xlarge instance type since
that has 3 hard drives we can utilize. The new
builder creates a partition on the front portion
of the drive to ensure that the test runs in a
reasonable amount of time.


dinatale2:
Still TODO:
- Move partition creation to bb-test-prepare.sh
- Possibly, zero out the partitions in bb-test-prepare.sh

There is also an issue when bb-test-zfstests.sh attempts to perform a module unload in its cleanup routine upon exit that doesn't appear to be related, but definitely needs looking into.

dinatale2:
Updated. Moved the partition creation to bb-test-prepare.sh.

dinatale2:
Updated. Attempting to zero out the partitions created in bb-test-prepare.sh. Also, generalized the logic for partitioning and zeroing.

dinatale2:
Updated. I think everything is all set for a performance builder. I need to look into the cleanup hang, but I'm sure that's not the fault of this pull request. We could wait until that is resolved to pull this in if you'd like.

dinatale2:
One last update, when running dd, use bs=1M for better performance.

behlendorf:
Merged!

PR #38, opened 2016-11-06
Add NFS and Samba server dependency for ZFS Test Suite

loli10K:
We need both NFS and Samba servers to run `zfs_share` and `zfs_unshare` scripts in the ZFS Test Suite.

Once this is merged i'll try enabling those tests in https://github.com/zfsonlinux/zfs/pull/5367.

I don't know all these distros very well, i hope i got the package names right.

### DIFF
--- a/scripts/bb-dependencies.sh
+++ b/scripts/bb-dependencies.sh
@@ -20,7 +20,7 @@ Amazon*)
 
     # Required utilities.
     $SUDO yum -y install git rpm-build wget curl bc fio acl sysstat \
-        mdadm lsscsi parted attr dbench watchdog ksh
+        mdadm lsscsi parted attr dbench watchdog ksh nfs-utils samba
     $SUDO yum -y install --enablerepo=epel cppcheck pax-utils
 
     # Required development libraries
@@ -38,7 +38,7 @@ CentOS*)
 
     # Required utilities.
     $SUDO yum -y install git rpm-build wget curl bc fio acl sysstat \
-        mdadm lsscsi parted attr dbench watchdog ksh
+        mdadm lsscsi parted attr dbench watchdog ksh nfs-utils samba
 
     # Required development libraries
     $SUDO yum -y install kernel-devel \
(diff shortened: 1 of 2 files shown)


loli10K:
@dinatale2 thanks, now i just need a portable way to enable/start NFS and Samba on every distro from the ZFS Test Suite ...

behlendorf:
@loli10K we should be able to get this patch rolled out when there's a lull in the automated testing.  If we treat starting the NFS/Samba services as part of configuring the environment before the test suite starts I'd suggest putting that logic in the [zfs-test-prepare.sh script](https://github.com/zfsonlinux/zfs-buildbot/blob/master/scripts/bb-test-prepare.sh).

loli10K:
@behlendorf i think that would work best, otherwise we would have to implement `setup_nfs_server` in https://github.com/zfsonlinux/zfs/blob/zfs-0.7.0-rc2/tests/zfs-tests/include/libtest.shlib#L1129 with a way to figure out which distro we're running on ... again.

behlendorf:
Sounds good.  Then it would be best to add that logic to this PR before we merge it.

behlendorf:
@loli10K merged and applied to the buildbot.  You should be able to open your PR against ZFS now which enables these tests.  We'll refine this change as needed.

loli10K:
@behlendorf thanks, i've already prototyped some changes, we should be able to pass 14 tests out of 16 (zfs_share+zfs_unshare) without too much work ... i'll try to push them by tomorrow.

Also, can i enable just those test in the runfile to speedup the process until i get the code right?

behlendorf:
That's great.  If you want to really speed up the test process you can modify the top level TEST file as follows and include it in your commit.  We'll just drop those changes in the final version.

* Disable the buildbot steps you want to skip:

TEST_ZTEST_SKIP="yes"
TEST_XFSTESTS_SKIP="yes"
...

* Add a new `share.run` to your commit and reference it in TEST_ZFSTESTS_RUNFILE

TEST_ZFSTESTS_RUNFILE="share.run"

* Request only the TEST builders test the PR by adding the following to the commit comment.

Requires-builders: test

PR #39, opened 2016-11-07
Requires-builder should accept the option 'all'

dinatale2:
When a commit message contains a Requires-builder
directive, check for 'all'. If 'all' is present, then
all available builders should be used.


behlendorf:
Merged as:

1166d7c Requires-builder should accept the option 'all'

Issue #40, opened 2016-12-14
priority of building


inkdot7:
I've noticed that the build system build commits in the order that they were pushed to the pull requests.  It also seems that all commits in a pull request are rebuilt, even if only a new commit is pushed at the end.

If possible, it would be nice to:
- First build all 'final' commits, i.e. the ones where currently the TEST builds are performed.
  Then single-commit pull requests are not so easily stalled by multi-commit requests.
- Prefer to build commits that have most other successfully built commits.
  Then code that fails style check and later fast builds will not hold the slower builders.

dinatale2:
@inkdot7 Thank you for opening this issue. These are all valid points. Below are my thoughts on each.

- Building all the 'final' commits may be doable. When creating a buildbot builder, it is possible to specify a function to choose the next build. Perhaps, that may be worth looking into?
- For optimizing which commits get built, we could do require that commits pass the STYLE builder before triggering other. We could create a "fast TEST" builder which can run a subset of the test suite and once that passes, we can then trigger the TEST builders.
- For the only building new commits, this is tracked by commit hash if I am not mistaken. I believe github issues a synchronize event via it's webhook. I believe it contains only the commit hashes that changed, but I could be wrong about that.

Would you be interested in attempting to tackle the above?

inkdot7:
@dinatale2 Thanks for the quick response.  I looked around in the code a bit:

- In `class CustomGitHubEventHandler` of `master/master.cfg`, there is `commits_cur` which gives the part number of a commit.  If adding this to the `change` dictionary makes it visible(?) to a `nextBuild` method of `class ZFSBuilderConfig` in `master/buildslaves.py` then reordering builds would be easy.  If the value is also set to `0` for `commits_cur == commits_num`, then it would be straight-forward for `nextBuild`: choose the first request with the lowest number: final commits gets built before others, and others are built such that the ends of long commit chains have to wait for shorter chains.

  I could try to make a patch/pull request.  But it would be untested, as I've no experience setting up a buildbot system.  Useful?

- Introducing a fast TEST would make the test builds finish later, which is not so nice.  As those have many builders, I've not felt any problems there.  Also, as the tests sometimes give spurious failures, it would not be nice if that prevents the other tests from running.

- That rebased commit chains are fully rebuilt makes sense.  It looks like extra rebuilds of unchanged commits when just one new is added is due to `class CustomGitHubEventHandler`, which adds all commits as changes when it gets told about a new update.

[...]

PR #41, opened 2016-12-15
Prioritise building the head (final) commit of each pull request.

inkdot7:
The last commits that make up long commit chains will be delayed when
shorter commit chains are pushed.  A large number of short commit chains
may starve them for long times.  Perhaps some increased priority for
waiting a long time (say 1 part unit per 5 h) should be employed.

Probably better to start out by letting chain-internal commits build in
original order, and just prioritise the head commits?

Note: this patch is untested!
Most notably: I am not sure if request.source.commit_part is a way that works
to access the commit_part item of the change request.
Feel free to change as needed!

See #40

### DIFF
--- a/master/buildslaves.py
+++ b/master/buildslaves.py
@@ -31,11 +31,27 @@ class ZFSBuilderConfig(util.BuilderConfig):
         # randomly choose among all our busy slaves
         return (random.choice(slaves) if slaves else None)
 
+    @staticmethod
+    def nextBuild(builder, requests):
+        min_part = None
+        min_request = None
+        for request in requests:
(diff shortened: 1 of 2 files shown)


dinatale2:
@inkdot7 The patch didn't work. I did some digging and found that adding fields to the change dictionary also doesn't work. Somewhere in the chain, the dictionary is used to create a change object and it only cares about specific fields.

I went ahead and decided to try some other solution using the Pull-request lines that the CustomGithubHandler puts into pull request comments. You can find it at [here](https://github.com/dinatale2/zfs-buildbot/tree/final-commit-priority). Thoughts on the alternative?

behlendorf:
Superseded by #42.

PR #42, opened 2016-12-20
Prioritize the last commit in a pull request

### LINKED ISSUE #40: priority of building

dinatale2:
Prioritize the last commit in a pull request or
recent mergings into master over other builds.
Parse out the part and total from the Pull-request
line in commit messages, if part is equal to total
then prioritize that build over others.

Closes #40

### DIFF
--- a/master/buildslaves.py
+++ b/master/buildslaves.py
@@ -31,11 +31,41 @@ class ZFSBuilderConfig(util.BuilderConfig):
         # randomly choose among all our busy slaves
         return (random.choice(slaves) if slaves else None)
 
-    def __init__(self, mergeRequests=False, nextSlave=None, **kwargs):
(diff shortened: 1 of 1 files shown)


behlendorf:
Merged as 7c2908dff7bad75f1008a425c99219d288a7af1a

PR #46, opened 2017-01-13
Add openzfs-tracking.sh script

behlendorf:
This script is used to dynamically generate the OpenZFS Commit
Tracking page at http://build.zfsonlinux.org/openzfs-tracking.html.

### DIFF
--- a/scripts/openzfs-exceptions.txt
+++ b/scripts/openzfs-exceptions.txt
@@ -1,0 +1,91 @@
+#
+# Commit exceptions used to explicitly reference a given Linux commit.
+# These exceptions are useful for a variety of reasons.
+#
+# - The OpenZFS commit isn't applicable to Linux (denoted by a -)
+# - The OpenZFS -> ZFS on Linux commit matching is unable to associate
+#   the related commits due to lack of information.
+# - The fix was merged to Linux prior to their being an OpenZFS issue.
+#
+# The exceptions are of the format:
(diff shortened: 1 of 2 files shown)


gmelikov:
I want to add some comments (links to wiki, etc) and table filter by commit status. Is it better to make new PR after this one or add my commit here?

behlendorf:
Improvements are welcome! If you have a better version of the script can you open a new PR with this commit and then your changes applied on top.   Thanks.

behlendorf:
@gmelikov actually let me just merge this since it's functional and useful.  This should make it easier for you to open a new PR with your various improvements.

jumbi77:
Thanks. The auto-gen. HTML page is cool. Maybe one can integrate the 'scripts/openzfs-exceptions.txt' entries with a different color and a note like 'The OpenZFS commit isn't applicable to Linux'. That way every commit/issue is tracked even it is not applicable to ZoL.

And a question: It is right that the following path '_usr/src/uts/common/fs/zfs_'  is twice entered in '_OPENZFS_PATHS='_ and '_usr/src/cmd/mdb/common/modules/zfs_' and '_usr/src/test/zfs-tests_' is missing? Just compare the "Directories of interest for ZFS" in the OpenZFS README.md

gmelikov:
@jumbi77 good idea, i'll try to implement it with another things!

PR #48, opened 2017-01-17
Improve openzfs-tracking.sh and add Tracking Exceptions

gmelikov:
- pagination and search box added
- exception comments added to html table
- minor bug fixes

Tracking Exceptions with questions from me:
- [4465](https://www.illumos.org/issues/4465) looks like autoregenerated, don't know what we can do with it.
- [6459](https://www.illumos.org/issues/6459) doesn't work properly with our codebase. Fast test - run `cstyle -pP` from openZFS, it begins to make warnings on all `foo()` without `{...` after it.

I think some exceptions are still there, ready to add if someone finds them, thanks.

[Example page](https://gmelikov.ru/zfs.html)


scripts/openzfs-exceptions.txt
```diff
@@ -24,6 +24,7 @@
 1974 -
 2130 -
 2665 32a9872
+2897 - 			already added
```
behlendorf: You can reference commit fb82700 for this.

scripts/openzfs-exceptions.txt
```diff
@@ -46,6 +47,7 @@
 4206 2820bc4
 4242 -
 4391 78e2739
+4465 - 			huge, looks like auto regenerated
```
behlendorf: All of the upstream OpenZFS documentation was converted to mdoc format.  In order to make it easier to update the documentation we need to pull in their updated mdoc pages and reapply the Linux specific changes/additions.

scripts/openzfs-exceptions.txt
```diff
@@ -86,10 +88,13 @@
 6418 131cc95
 6421 ca0bf58
 6434 472e7c6
+6459 - 			changes don't work properly with ZoL codebase
```
behlendorf: These changes should work with ZoL.  However, it's entirely possible, even likely, that there are places in the ZoL code base which are Linux specified which need to be updated.

gmelikov: I already ported it ([link](https://github.com/gmelikov/zfs/blob/OpenZFS-6459/scripts/cstyle.pl)).

Example of strange output: 
```
./module/zfs/vdev_raidz_math_ssse3.c: 423: stuff after {
./module/zfs/vdev_raidz_math_ssse3.c: 424: non-continuation indented 4 spaces
```

There are good warnings too. I'll add full out here.
[out.txt](https://github.com/zfsonlinux/zfs-buildbot/files/712529/out.txt)

scripts/openzfs-exceptions.txt
gmelikov: Thanks, i'll try to work on it later.

scripts/openzfs-exceptions.txt
behlendorf: It appears that most (maybe all) of the resulting warnings are in Linux specific areas.  So we'll want to address them which pulling in the cstyle.pl updates.

All of those `vdev_raidz_math_ssse3.c` warnings should be resolvable with the following small change.  I suspect we'll need to make the same change in other places for the other optimization raidz implementations.

```diff
diff --git a/module/zfs/vdev_raidz_math_ssse3.c b/module/zfs/vdev_raidz_math_sss
index cebb0fe..1742115 100644
--- a/module/zfs/vdev_raidz_math_ssse3.c
+++ b/module/zfs/vdev_raidz_math_ssse3.c
@@ -419,7 +419,8 @@ const raidz_impl_ops_t vdev_raidz_ssse3_impl = {
 #if defined(HAVE_SSSE3) || defined(HAVE_AVX2) || defined(HAVE_AVX512BW)
 
 const uint8_t
-__attribute__((aligned(256))) gf_clmul_mod_lt[4*256][16] = {
+__attribute__((aligned(256))) gf_clmul_mod_lt[4*256][16] =
+{
        { 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
            0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00  },
        { 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
```

There may be other places where complicated macros are used where you'll need to disable the style checking entirely.  You can do this by wrapping the block with the following comments.

```
/* BEGIN CSTYLED */
...
/* END CSTYLED */
```

gmelikov: I'll make a PR to assign it as WIP and push separate commit with style changes. Thanks for clarification!

scripts/openzfs-tracking.sh
```diff
@@ -221,13 +222,17 @@ do
...
+	EXCEPTION=$(grep -E "^$OPENZFS_ISSUE\s" "$ZFSONLINUX_EXCEPTIONS")
 	if [ -n "$EXCEPTION" ]; then
 		EXCEPTION_HASH=$(echo $EXCEPTION | cut -f2 -d' ')
+		EXCEPTION_COMMENT=$(echo $EXCEPTION | cut -d' ' -f3-)
 		if [ "$EXCEPTION_HASH" == "-" ]; then
 			ZFSONLINUX_HASH="-"
 			ZFSONLINUX_STATUS=$STATUS_NONAPPLICABLE
 			ZFSONLINUX_STATUS_TEXT=$STATUS_NONAPPLICABLE_TEXT
+			if [ -n "$EXCEPTION_COMMENT" ]; then
+			OPENZFS_DESC="$OPENZFS_DESC</br><b>comment: $EXCEPTION_COMMENT</b>"
```
behlendorf: Missing tab.  But I like the idea of making the comment visible in the web page.

gmelikov: Thanks, fixed. Your idea is great, but the question is how will we do that. Maybe we can open PR with WIP label and parse special formatted string from it's text? We can color them differently by this label too.

behlendorf: That would work.  But I'm not keen on having 70 odd PRs open which I can't close.

Alternately, I was thinking we could just add a page on the wiki and scrape it for the exception information rather than using the checked in version.  We could just use `curl` to grab the raw page from Github.

hat would allow project members to easily update and maintain it.  And it would still be simple to parse.  We could just extend the existing format slightly so a `!` in the second column means the commit is applicable but not applied for the reason described in the comment.

gmelikov: That's great! I think we can move exceptions to wiki too, that would be easier than make PRs here for every exception in future. 

Yes, merge it, thanks. I'll open the new one.

Issue #50, opened 2017-01-25
Automate OpenZFS patch porting


dinatale2:
We recently discovered that it is possible to cherry-pick patches directly from the OpenZFS tree. Git will automatically attempt to find the right file to apply a hunk to if it is given enough time to process a change.

To allow git the time to search, you need the following in your git config
```
[merge]
    renameLimit = 999999
```
Now, in order to cherry-pick, clone your fork of zfs and add both zfsonlinux/zfs and openzfs/openzfs as remotes. Perform a `git fetch --all`. Now you can perform a `git cherry-pick <OpenZFS commit>` to port patches to ZoL.

It would be nice to have a script that would attempt to perform the above process automatically given an OpenZFS commit.

dinatale2:
@gmelikov you might be interested in this

gmelikov:
Thanks, i thought about that scenario too. I'll try to make a script for it.

gmelikov:
@dinatale2 thanks for initial method!
I alreadt made a prototype, it works smoothly. I'll make a PR after massive test.

gmelikov:
Most of my last PRs are made by script, but unfourtunately these are all that can be merged automatically for now.

One major problem - new files merge with OpenZFS file structure, example - https://github.com/gmelikov/zfs/commit/df3e48b064a553fa0fc9178d2a71fc57008edb0c

I'll beautify the script and make a PR later.

dinatale2:
This was taken care of in #51. Thanks @gmelikov!

PR #51, opened 2017-01-25
Add OpenZFS auto merge script

gmelikov:
This script can:
- iterate all unmerged commits and try to merge it
- create branch with changes on success
- check for compile and cstyle errors
- generate lists with commits in different state (compile errors, style errors, merge errors)

`openzfs-tracking.sh ` now can generate text files with unmerged OpenZFS commits, if add `-c` parameter.
We can post them on website too for easier auto merge process.

This is a fast written draft, welcome any thoughts. I think it will be better to wait some time before merge it.

I tested it, you can see last PRs from my `autoport` branches.

Unfourtunately, there is no suitable to auto merge commit, you can add https://github.com/openzfs/openzfs/commit/9185393 to hashes file manually.

Based on #50.


gmelikov:
Add manual mode:

- default - iterate commits and try to merge without manual interruption
- `-c` - merge specified commit (by hash) without undo on failure. Makes branch with merge conflicts.
- `-g` - generate commit description and `git commit --amend` it to last commit (by commit hash, work in created by `-c` branch)

Now i can do:
- `./openzfs-merge.sh -d /path/to/zfs -c 40510e8` to begin cherry-picking
- edit any found conflicts by `git mergetool`
- `git add/rm`
- `git commit` with any description
-`./openzfs-merge.sh -d /path/to/zfs -g 40510e8` to end with description

Maybe it's an overkill, but it makes porting easier for me.

behlendorf:
@gmelikov thanks, this looks really useful and your manual mode looks ideal for the use case I have in mind.  Once we're largely caught up with the OpenZFS tree we're going to want to automate this process as much as possible.  As you know this won't work for all commits, but there will be a subset of commits which do apply cleanly.

Using your manual mode we'd love to be able to script some version the following which runs when a new commit gets merged to OpenZFS.

- Cherry pick the commit hash on to a new branch
- Iff it applies cleanly run the style and compile check
- Iff the checks pass open a new PR for automated testing

Since you've already implemented the majority of this I think the only missing bit is the ability to open a new PR.  The good news here is that there's already an existing utility provided by GitHub which does this called [hub](https://hub.github.com/).  It's a wrapper which extends to git command to support things like opening PRs from the command line.  If you can add this functionality to the script we should have everything we need.

Thanks for working on this!

gmelikov:
@behlendorf thanks for straightforward usecase. I'll add this functionality little later, because while perfecting script i'm ready to check by myself. But it's the best approach, i agree.

Main problems with auto merge:
- different paths to files - maybe i'll add symlink for main cases
- little differences with line order (for ex. copyright comments) - the only decision is to make as identical as possible on manual merge. I already began this in https://github.com/zfsonlinux/zfs/pull/5670
- after all we have a man pages problem (different formats)

behlendorf:
> different paths to files - maybe i'll add symlinks for main cases

This _should_ be handled by git in the majority of cases.  There are going to be a couple of files which it can't match because they've been either rewritten entirely for Linux (vdev_disk.c, zvol.c, zfs_ctldir.c) or currently only exist under Linux (zpl_*.c, vdev_raidz_math*.c).  Have you noticed other files which can't me matched?

> little differences with line order (for ex. copyright comments) - the only decision is to make as identical as possible on manual merge. 

Agreed, that's the best we can do.  I'm not opposed to applying patches to the Linux port which are just designed to minimize these tiny differences.

> after all we have a man pages problem

Yes, we should grab the rewritten mdoc pages from Illumos as a starting point and then re-apply the Linux specific sections we've added.  This would also be a good time to feed trivial fixes for spelling mistakes and such back upstream to OpenZFS.

scripts/openzfs-merge.sh
```diff
@@ -0,0 +1,231 @@
...
+	Auto merge OpenZFS commits to ZFS on Linux git
+	repositories.
+	Result - git branch with name 'autoport-oz#issue'
+
+OPTIONS:
+	-h		Show this message
+	-d directory	Git repo with openzfs and zfsonlinux remotes
+	-i commits.txt	File with OpenZFS commit hashes to merge
+	(one hash per row)
+	-i commit hash	Prepare branch and try to merge this commit by hash,
```
dinatale2: I think you meant `-c` here.

scripts/openzfs-merge.sh
```diff
@@ -0,0 +1,231 @@
...
+	git branch -D "autoport-oz$OPENZFS_ISSUE"
+	rm "$TMP_FILE"
+}
+
+prepare_git() {
+	cd "$REPOSITORY_PATH"
+	rm "$TMP_FILE"
+	git checkout master
+	git fetch --all
+	git log --remotes=openzfs/master --format=%B -n 1 $OPENZFS_COMMIT > "$TMP_FILE"
```
dinatale2: Is `--remotes` necessary here? The commit hash should be sufficient to find the log entry.

scripts/openzfs-merge.sh
```diff
@@ -0,0 +1,231 @@
...
+#       email = mail@gmelikov.ru
+#       name = George Melikov
+#
+# Copyright (c) 2016 George Melikov. All rights reserved.
+#
+
+SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+REPOSITORY_PATH='.'
+TMP_FILE='/tmp/gittmpmessage.txt'
```
dinatale2: I think you are using a file because it is easier to work with? If so, then this fine.

scripts/openzfs-merge.sh
```diff
@@ -0,0 +1,231 @@
...
+	    printf 'compilation failed' >&2
+	    echo $OPENZFS_COMMIT >> "$EXCEPTIONS_COMPILE"
+	    COUNT_COMPILE=$(($COUNT_COMPILE+1))
+	    LIST_COMPILE="$LIST_COMPILE
+	    autoport-oz$OPENZFS_ISSUE"
+	    ERR=1
+	fi
+	
+	echo -e "${LGREEN} Cstyle... ${NORMAL}"
+	if ! make cstyle; then
```
dinatale2: I'm not sure if you should fail if cstyle fails. The patch may work and would involve minor tweaking to address the style problems. Thoughts?

scripts/openzfs-tracking.sh
```diff
@@ -81,6 +81,8 @@ OPTIONS:
 	-h		Show this message
 	-d directory	Git repo with openzfs and zfsonlinux remotes
 	-e exceptions	Exception file (using ZoL wiki if not specified)
+	-c file.txt		Write OpenZFS unmerged commits' hashes to file,
```
dinatale2: Tabbing/Spacing needs to be fixed here.

scripts/openzfs-tracking.sh
```diff
@@ -234,7 +239,7 @@ do
 
 	# Match issue against any open pull requests.
 	ZFSONLINUX_PR=$(echo $ZFSONLINUX_PRS | jq -r ".[] | select(.title | \
-	    contains(\"OpenZFS $OPENZFS_ISSUE\")) | { html_url: .html_url }" | \
+	    contains(\"OpenZFS $OPENZFS_ISSUE \")) | { html_url: .html_url }" | \
```
dinatale2: A rebase should get rid of this change.

scripts/openzfs-merge.sh
```diff
@@ -0,0 +1,231 @@
...
+# This script tries to merge OpenZFS commits to ZoL.
+#
+# Instruction:
+# 
+# Repository setup must be similar with openzfs-tracking.sh
+# requirements.
+#
+# Repository setup for valid compilation check:
+#	sh autogen.sh
+#	./configure --enable-debug
```
dinatale2: Wouldn't it be safe to run `sh autogen.sh` and `./configure` for each commit? Is it because Makefile.in's don't often change? There are other .in files as well.

scripts/openzfs-merge.sh
```diff
@@ -0,0 +1,231 @@
...
+	leaves branch for manual merge if not successfull
+	-g commit hash	Generate commit description for existing commit in
+	branch generated by -i parameter
+
+EOF
+}
+
+clean_unmerged() {
+	git cherry-pick --abort
+	git checkout master
```
dinatale2: It may be worth documenting that this script should be used on a clone where the origin is a fork of ZFS on Linux that has a remote to OpenZFS.

scripts/openzfs-merge.sh
```diff
@@ -0,0 +1,231 @@
...
+	git branch -D "autoport-oz$OPENZFS_ISSUE"
+	rm "$TMP_FILE"
```
dinatale2: You may want to run a `git clean -xdf`. This would require you to rerun `sh autogen.sh` and `./configure`.

scripts/openzfs-merge.sh
behlendorf: It may be appropriate to add a flag to the script to control this behavior.

scripts/openzfs-merge.sh
```diff
@@ -0,0 +1,231 @@
...
+}
+
+merge() {
+	ERR=0
+	
+	prepare_git
+	
+	echo -e "${LGREEN} - OpenZFS issue #$OPENZFS_ISSUE $OPENZFS_COMMIT ${NORMAL}"
+	echo -e "${LGREEN} Checkout new branch ${NORMAL}"
+	git branch -D "autoport-oz$OPENZFS_ISSUE"
```
dinatale2: You may want to change this to `git branch -D "autoport-oz$OPENZFS_ISSUE" > /dev/null` because it displays an error if the branch doesn't exist.

scripts/openzfs-merge.sh
```diff
@@ -0,0 +1,231 @@
...
+	OPENZFS_ISSUE=$(grep -oP '^[^0-9]*\K[0-9]+' -m 1 "$TMP_FILE")
+}
+
+generate_desc() {
+	USER_NAME=$(git config user.name)
+	USER_MAIL=$(git config user.email)
+	OPENZFS_COMMIT_AUTHOR=$(git log --format="%aN <%aE>" --remotes=openzfs/master -n 1 $OPENZFS_COMMIT)
+	
+	sed -i '1s/^/OpenZFS /' "$TMP_FILE"
+	sed -i 's/ / - /2' "$TMP_FILE"
```
dinatale2: I believe this string replace intends to add a dash after the issue number, but it ends up putting dashes after each `Reviewed by:`. Also, it's not spaced correctly. Here's what I came up with:

```
generate_desc() {
        USER_NAME=$(git config user.name)
        USER_MAIL=$(git config user.email)
        OPENZFS_COMMIT_AUTHOR=$(git log --format="%aN <%aE>" --remotes=openzfs/master -n 1 $OPENZFS_COMMIT)
        OPENZFS_ISSUE=$(grep -oP '^[^0-9]*\K[0-9]+' -m 1 "$TMP_FILE")

        sed -i "1s/^$OPENZFS_ISSUE/OpenZFS $OPENZFS_ISSUE -/" "$TMP_FILE"
        sed -i "1 a Authored by: $OPENZFS_COMMIT_AUTHOR" "$TMP_FILE"
        sed -i '/^$/d' "$TMP_FILE"
        sed -i "/^OpenZFS $OPENZFS_ISSUE.*$/G" "$TMP_FILE"

        echo 'Ported-by: '$USER_NAME' <'$USER_MAIL'>' >> "$TMP_FILE"
        echo '' >> "$TMP_FILE"
        echo 'OpenZFS-issue: https://www.illumos.org/issues/'$OPENZFS_ISSUE >> "$TMP_FILE"
        echo 'OpenZFS-commit: https://github.com/openzfs/openzfs/commit/'$OPENZFS_COMMIT >> "$TMP_FILE"
}
```

scripts/openzfs-merge.sh
gmelikov: I think in this way: patch have to be edited -> fail, overwise success.
Yes, it's easy to add a flag, i'll add it.

scripts/openzfs-merge.sh
gmelikov: Thanks, fixed this problem.

scripts/openzfs-merge.sh
gmelikov: It will we safer, but much more slower.

I didn't have any issues with it yet.

scripts/openzfs-merge.sh
gmelikov: Yes, it was just easier for me.

scripts/openzfs-merge.sh
gmelikov: There are already these lines:
```
# Repository setup must be similar with openzfs-tracking.sh
# requirements.
```
openzfs-tracking.sh already has there requirements.

scripts/openzfs-merge.sh
gmelikov: Maybe, but this variant is faster and it won't remove untracked files.

scripts/openzfs-merge.sh
gmelikov: It's just for safety, we use short hashes.

gmelikov:
@behlendorf just have examples for different path:
commit https://github.com/openzfs/openzfs/commit/39cddb1
```
	deleted by us:   usr/src/uts/common/fs/zfs/sys/vdev_disk.h
	deleted by us:   usr/src/uts/common/fs/zfs/vdev_disk.c
	deleted by us:   usr/src/uts/common/os/devcfg.c
```

scripts/openzfs-merge.sh
dinatale2: Ah, I missed that. Thanks.

behlendorf:
So in this case the `vdev_disk.*` files have been effectively rewritten from scratch for Linux and `devcfg.c` isn't part of ZFS just illumos.  I'll update the exceptions page with a comment describing the situation for this commit.

behlendorf:
@gmelikov let's get this merged as an initial version of the script.  Do you have any additional fixes yould like to make first?

gmelikov:
@behlendorf i don't have any for now, so let's merge it. Thanks for review!

PR #55, opened 2017-02-01
Introduce a flag to control cstyle failures in auto merge script

dinatale2:
Introduce -s flag to control if a cstyle failure should cause
the merge process to fail. Corrected some style as well.


dinatale2:
@gmelikov would you mind taking a look?

PR #56, opened 2017-02-01
Fix TEST_ZFSTESTS_DISKSIZE environment variable

behlendorf:
There was a typo resulting in the TEST_ZFSTESTS_DISKSIZE disk
size variable being ignored when sourced from the environment.


behlendorf:
Thanks!

PR #57, opened 2017-02-05
Change maxTime for ziltest

gmelikov:
It's needed for https://github.com/zfsonlinux/zfs/pull/5679


behlendorf:
A normal run of ziltest should only take a few seconds to run.  We shouldn't need to increase the timeout to get this to pass.  There's probably an issue with the test case which needs to be investigated.

http://build.zfsonlinux.org/builders/Amazon%202015.09%20x86_64%20Release%20%28TEST%29/builds/1301/steps/shell_6/logs/stdio

gmelikov:
@behlendorf thanks for the information, i'll try to investigate it further.

Sorry for the false PR.

PR #60, opened 2017-02-14
openzfs-merge.sh: add -p flag - push branch to github

gmelikov:
Small change allows to push branch with merged patch to github with less work.


gmelikov:
@dinatale2 you're right, but I added this flag to easily push commits which can't be automatically merged (for ex. i made it to use with `-g` flag). I added comment that the new flag forces it.

scripts/openzfs-merge.sh
```diff
@@ -106,6 +109,9 @@ generate_desc() {
 #add description to commit
 add_desc_to_commit() {
 	git commit --amend -F "$TMP_FILE"
+	if [ $PUSH_NEW_BRANCH -ne 0 ]; then
```
dinatale2: I would suggest moving this out into the merge function. You can do something like
```
if [ $PUSH_NEW_BRANCH -ne 0 ] || [ $ERR -eq 0 ]; then
# push branch here
fi
```

Then you can remove the git push from the existing ERR check.

gmelikov: I made it for the case with manual merge, so after manual merge is done I use -g key to generate a valid description, so if it will be in merge function it won't run. But I understood you, will move it and do it more cleanly.

gmelikov: Made a function for it.

PR #66, opened 2017-03-20
zfstests buildstep should fail if zfs-tests.sh fails

dinatale2:
bb-tests-zfstests.sh was not accounting for a zfs-tests.sh
failure. Intrduce logic to handle cases where zfs-tests.sh
fails and ensure an error code is returned.

### DIFF
--- a/scripts/bb-test-zfstests.sh
+++ b/scripts/bb-test-zfstests.sh
@@ -68,9 +68,12 @@ $ZFSTESTS $TEST_ZFSTESTS_OPTIONS \
     -d $TEST_ZFSTESTS_DIR \
     -s $TEST_ZFSTESTS_DISKSIZE \
     -r $TEST_ZFSTESTS_RUNFILE
+RESULT=$?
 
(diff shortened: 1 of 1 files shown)


scripts/bb-test-zfstests.sh
```diff
@@ -68,9 +68,10 @@ $ZFSTESTS $TEST_ZFSTESTS_OPTIONS \
     -d $TEST_ZFSTESTS_DIR \
     -s $TEST_ZFSTESTS_DISKSIZE \
     -r $TEST_ZFSTESTS_RUNFILE
+RESULT=$?
 
 # FAILURE: One or more test cases failed.
-grep "\[FAIL\]" log && RESULT=1
+[ $RESULT -eq 0 ] && grep "\[FAIL\]" log && RESULT=1
```
behlendorf: This would be clearer as a normal `if` conditional.

dinatale2:
Refreshed it so it's a normal if condition.

PR #67, opened 2017-03-28
Silence removal of git message temp file

dinatale2:
The temp file won't always exist, so the error
can be ignored. Pipe rm output to /dev/null.

### DIFF
--- a/scripts/openzfs-merge.sh
+++ b/scripts/openzfs-merge.sh
@@ -80,12 +80,12 @@ clean_unmerged() {
 	git cherry-pick --abort
 	git checkout master
 	git branch -D "autoport-oz$OPENZFS_ISSUE" > /dev/null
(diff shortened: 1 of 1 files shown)


scripts/openzfs-merge.sh
```diff
@@ -80,12 +80,12 @@ clean_unmerged() {
 	git cherry-pick --abort
 	git checkout master
 	git branch -D "autoport-oz$OPENZFS_ISSUE" > /dev/null
-	rm "$TMP_FILE"
+	rm "$TMP_FILE" > /dev/null
```
behlendorf: `rm -f "$TMP_FILE"` is generally a cleaner way to handle this.  It won't generate a warning even if the file is missing.

dinatale2:
@behlendorf Corrected the `rm` calls to use `-f`.

PR #70, opened 2017-03-31
Add rng-tools to dependencies

crawfxrd:
rng-tools will be used to ensure entropy is available for ZFS tests that use encryption (zfsonlinux/zfs#5769).


dinatale2:
@crawfxrd thanks!

PR #71, opened 2017-04-02
openzfs-tracking.sh - fix pending Linux commit generation

gmelikov:
Link title of pending commit had `PR-` text without issue number, fixed (example - OpenZFS 7290).


dinatale2:
@gmelikov Thanks!

PR #75, opened 2017-05-19
KILLED tests should result in a warning

behlendorf:
Test cases which were KILLED and not FAILED would result
in the over status still being SUCCESS.  Update the script
so a KILLED test results in a WARNINGS status (2).  Any
FAILED test will still resulted in a FAILURE overall status.


behlendorf:
Here's a [recent example](http://build.zfsonlinux.org/builders/Amazon%202015.09%20x86_64%20Release%20%28TEST%29/builds/2428).

PR #76, opened 2017-06-05
Detect when OOM-killer activity occurs

dinatale2:
Print a message if OOM-killer activity is detected
in console logs during zfstests, xfstests, and zfsstress.

Blacklist the buildslave process from being OOM-killed.

### DIFF
--- a/scripts/bb-test-prepare.sh
+++ b/scripts/bb-test-prepare.sh
@@ -22,6 +22,15 @@ SUDO="sudo -E"
 
 set -x
 
+# Attempt to set oom_score_adj for buildslave to prevent
+# it from being targeted by the oom-killer
+if test -f "$BB_DIR/twistd.pid"; then
+    pid=$(cat "$BB_DIR/twistd.pid")
+    if test -f "/proc/${pid}/oom_score_adj"; then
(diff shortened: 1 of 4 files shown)


scripts/bb-test-prepare.sh
```diff
@@ -22,6 +22,13 @@ SUDO="sudo -E"
 
 set -x
 
+# Attempt to blacklist the buildslave to prevent it
+# from being targeted by the oom-killer
+if test -d "$BB_DIR"; then
```
behlendorf: Testing for the existence of the file here rather than the directory would probably make more sense.  It would also be good to check for the existence of the proc file since I'm not sure which kernel version added this .  This can all be best effort.

scripts/bb-test-prepare.sh
```diff
@@ -22,6 +22,13 @@ SUDO="sudo -E"
 
 set -x
 
+# Attempt to blacklist the buildslave to prevent it
+# from being targeted by the oom-killer
```
behlendorf: How about something more along the lines of, "Attempting to set oom_score_adj for buildslave to prevent...".

dinatale2:
@behlendorf updated the commit message and calling `sudo` when attempting to update `oom_score_adj` for permission reasons.

PR #82, opened 2017-08-15
Make BUILD builders obey build properties

dinatale2:
BUILD builders need to obey a build properties
assigned to them by buildbot. These builders were
running with default configure options set in the
bb-build-zfs.sh script.


dinatale2:
@behlendorf I refreshed this patch.

PR #85, opened 2017-08-22
Allow build property overrides in commit messages

dinatale2:
Allow for users to control if and how zfs, spl, and lustre are
built in commit messages. If an override is provided, they will
replace the builder specified defaults.

### DIFF
--- a/master/master.cfg
+++ b/master/master.cfg
@@ -6,6 +6,7 @@ from twisted.python import log
 from password import *
 from buildslaves import *
 from buildbot.buildslave import BuildSlave
+from buildbot.plugins import util
 
 bb_slave_port = 9989
 bb_try_port = 8033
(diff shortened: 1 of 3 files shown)


dinatale2:
@behlendorf refreshed with what I think is close, if not, the final version.

master/master.cfg
```diff
@@ -154,22 +191,26 @@ build_factory.addStep(Git(repourl=spl_repo, workdir="build/spl",
...
+    description=["cloning"], descriptionDone=["cloned"],
     doStepIf = do_step_build_lustre,
-    hideStepIf=lambda results, s: results==SKIPPED,
-    description=["cloning"], descriptionDone=["cloned"]))
+    hideStepIf=lambda results, s: results==SKIPPED))
 build_factory.addStep(ShellCommand(
     workdir="build/linux", env={'PATH' : bin_path,
-        'LINUX_BUILTIN' : util.Interpolate('%(prop:builtin:-no)s') },
+        'LINUX_BUILTIN' : util.Interpolate('%(kw:p)s', p=get_builtin) },
     command=["runurl", bb_url + "bb-build-linux.sh"],
```
behlendorf: It looks like `bb-build-linux.sh` is also going to need an update such that it doesn't always assume the `../spl/` directory exists.

dinatale2: Fixed.

PR #87, opened 2017-09-01
Added dependency on openssl-devel

tcaputi:
openssl-devel includes a PBKDF2 implementation required
by ZoL PR #6595.

Signed-off-by: Tom Caputi <tcaputi@datto.com>


dinatale2:
Thanks! Merged.

PR #92, opened 2017-09-12
Force EC2 instances to shutdown after 8 hours

dinatale2:
We have cases of EC2 instances being stranded by buildbot.
Introduce a timed shutdown so TEST slaves don't persist
for large amounts of time.

### DIFF
--- a/scripts/bb-test-prepare.sh
+++ b/scripts/bb-test-prepare.sh
@@ -219,4 +219,16 @@ case "$BB_NAME" in
         ;;
 esac
 
+# Schedule a shutdown for all distros other than CentOS 6, Ubuntu 14.04,
+# and Amazon based distros
+case "$BB_NAME" in
+    Amazon*|CentOS-6*|Ubuntu-14.04*)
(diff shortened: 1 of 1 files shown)


dinatale2:
Want me to go ahead and merge it as is? I was going to test it tomorrow.

dinatale2:
@gmelikov @behlendorf went ahead and refreshed this. I'm now calling `shutdown` in `bb-test-prepare.sh` (seemed appropriate) and I'm backgrounding `shutdown` so it doesn't cause bb to stall.

behlendorf:
Seems reasonable, but we'll want to test it.

dinatale2:
So `shutdown` behaves differently on CentOS 6 than all of the other distros we support on buildbot. On all the other distros, calling `shutdown +x &` causes a scheduled shutdown. CentOS 6 requires return to be pressed.

dinatale2:
@behlendorf @gmelikov I've updated this. I only now call `shutdown` on non-CentOS 6 and non-Ubuntu 14.04 builders.

PR #96, opened 2017-09-15
Add support for integration with codecov.io

behlendorf:
* Added support for in-tree testing by adding an 'install' property.
  This controls whether packages should be installed or not, and if
  the bb-test-* scripts use should using the installed zfs utilities
  or those in-tree.  By running everything in-tree the user space
  gcov profiling files are kept with the source.  The kernel space
  profiling files are copied out of debugfs.

* The bb-build-packages.sh and bb-build-tree.sh script were
  unified in to bb-build.sh.  The INSTALL_METHOD environment
  variable is used to control if packages are build and installed.
  By default only the source is built.

* Added bb-test-cleanup.sh script which is called from a new buildbot
  cleanup step.  It responsible for doing any post processing after
  all testing has finished.  This currently entails running gcov
  and uploading the results to codecov.io.

* The kmemleak builder renamed the coverage builder

* Replaced SUDO environment variable with 'sudo -E'.

* All bb-test-* scripts now default to the build/tests directory.

* Minor style consistency changes.


scripts/bb-test-cleanup.sh
```diff
@@ -0,0 +1,106 @@
...
+{
+    pushd "${ZFS_BUILD}" >/dev/null
+    curl -s https://codecov.io/bash | bash -s - \
+        -c -Z -X gcov -X py -X xcode \
+        -n "$BUILDER_NAME" \
+        -b "$BUILD_NUMBER" \
+        -C "$ZFS_REVISION" \
+        -B "$BASE_BRANCH" \
+        -P "$PR_NUMBER" \
+        -F "$1"
```
prakashsurya: this won't hurt anything, but we're not using the "$1" parameter in the current iteration of this script.. I was using it when testing an older version.. thus, we can remove this to avoid any confusion.. leaving it won't hurt, it's just dead code.

PR #101, opened 2017-09-19
Clone repos before builds on TEST builders

dinatale2:
Clone all repos necessary for building before
building and testing on TEST builders. This will
avoid TEST builders from doing work if a stale
commit is present.


dinatale2:
@behlendorf updated

PR #112, opened 2017-09-29
codecov: log static HTML coverage reports

prakashsurya:
NOTE: This patch hasn't been tested.

This patch attempts to collect and log the static HTML coverage reports
that are generated via `make code-coverage-capture`, such that they can
be easily downloaded using the Buildbot UI, and viewed on a developer's
local workstation.

### DIFF
--- a/master/master.cfg
+++ b/master/master.cfg
@@ -437,6 +437,8 @@ test_factory.addStep(ShellCommand(
     command=["runurl", bb_url + "bb-test-cleanup.sh"],
     haltOnFailure=False, maxTime=1800, sigtermTime=30, logEnviron=False,
     lazylogfiles=True,
+    logfiles={"coverage-user"   : { "filename" : "coverage-user.tar.xz",   "follow" : False },
+              "coverage-kernel" : { "filename" : "coverage-kernel.tar.xz", "follow" : False }},
     decodeRC={0 : SUCCESS, 1 : FAILURE, 2 : WARNINGS, 3 : SKIPPED },
     description=["analysis"], descriptionDone=["analysis"],
     hideStepIf=lambda results, s: results==SKIPPED))
--- a/scripts/bb-test-cleanup.sh
+++ b/scripts/bb-test-cleanup.sh
@@ -11,6 +11,7 @@ else
     exit 1
 fi
 
+WORKDIR=$(readlink -f .)
 GCOV_KERNEL="/sys/kernel/debug/gcov"
 ZFS_BUILD=$(readlink -f ../zfs)
 
@@ -79,6 +80,14 @@ function upload_codecov_report_with_flag
         exit 1
     fi
 
+    #
+    # This configures the directory name that will be used for the HTML
+    # report that will be generated. Since we'll create multiple
+    # different reports, and we want to capture and log all of them
+    # via Buildbot, we use a different directory name for each one.
+    #
+    export CODE_COVERAGE_OUTPUT_DIRECTORY="coverage-$1"
+
     make code-coverage-capture
     curl -s https://codecov.io/bash | bash -s - \
         -c -Z -X gcov -X py -X xcode \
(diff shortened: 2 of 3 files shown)


master/master.cfg
```diff
@@ -550,7 +550,9 @@ perf_factory.addStep(ShellCommand(
     command=["runurl", bb_url + "bb-test-cleanup.sh"],
     haltOnFailure=False, maxTime=120, sigtermTime=30, logEnviron=False,
     lazylogfiles=True,
-    logfiles={"test"     : { "filename" : "TEST",        "follow" : False }},
+    logfiles={"test"            : { "filename" : "TEST",                   "follow" : False },
+              "coverage-user"   : { "filename" : "coverage-user.tar.gz",   "follow" : True  },
+              "coverage-kernel" : { "filename" : "coverage-kernel.tar.gz", "follow" : True  }},
```
behlendorf: For the tarballs you're going to want to set `"follow" : False, "lazylogfiles" : True`, the options are describe [here](http://docs.buildbot.net/0.8.1/Using-ShellCommands.html).  This way the links to the logs should only appear for the coverage builder.

You should also add this to the `test_factory` and not the `perf_factory`.

master/master.cfg
```diff
@@ -550,7 +550,9 @@ perf_factory.addStep(ShellCommand(
     command=["runurl", bb_url + "bb-test-cleanup.sh"],
     haltOnFailure=False, maxTime=120, sigtermTime=30, logEnviron=False,
     lazylogfiles=True,
-    logfiles={"test"     : { "filename" : "TEST",        "follow" : False }},
+    logfiles={"test"            : { "filename" : "TEST",                   "follow" : False },
+              "coverage-user"   : { "filename" : "coverage-user.tar.xz",   "follow" : True  },
+              "coverage-kernel" : { "filename" : "coverage-kernel.tar.xz", "follow" : True  }},
```
behlendorf: For the tarballs you're going to want to set `"follow" : False, "lazylogfiles" : True`, the options are describe [here](http://docs.buildbot.net/0.8.1/Using-ShellCommands.html).  This way the links to the logs should only appear for the coverage builder.

You should also add this to the `test_factory` and not the `perf_factory`.

scripts/bb-test-cleanup.sh
```diff
@@ -87,6 +97,17 @@ function upload_codecov_report_with_flag
...
+    #
+    # The compressed tarball will be collected and stored on the
+    # Buildbot "master". To try and reduce the amount of storage
+    # required to store these coverage reports, we do our best to
+    # compress the tarball as much possible. We measured xz to
+    # result in a file that's roughly half the size when compared
+    # to the output when using gzip; no other algorithms were tested.
+    #
+    tar -cf - "$CODE_COVERAGE_OUTPUT_DIRECTORY" | \
+        xz -9e "${WORKDIR}/${CODE_COVERAGE_OUTPUT_DIRECTORY}.tar.xz"
```
behlendorf: Let's make this run conditionally on a `TEST_CODE_COVERAGE_HTML` variable.  That way we can leave it disabled by default while we work on the issues with PRs.

dinatale2: Does `xz` require the full path as well considering it's being created in the current working directory?

dinatale2:
@prakashsurya Out of curiosity, how big is the tarballs?

prakashsurya:
@behlendorf I'm not sure I made the changes to the `master.cfg` file correctly; can you check my update?

prakashsurya:
@dinatale2 I ran zloop for an hour, and then ZTS and generated a coverage report for userspace only.. using the HTML files for that report, here's what I saw on my VM:
```
$ time sh -c 'tar -cf - zfs-0.7.0-coverage | xz -9e > zfs-0.7.0-coverage.tar.xz'

real    0m44.717s
user    0m44.540s
sys     0m0.230s

$ du -sh zfs-0.7.0-coverage.tar.xz
2.2M    zfs-0.7.0-coverage.tar.xz
```
so that's 2.2 MB; I'd expect the kernel report to be similar in size.

prakashsurya:
```
--- a/master/master.cfg
+++ b/master/master.cfg
@@ -437,8 +437,8 @@ test_factory.addStep(ShellCommand(
     command=["runurl", bb_url + "bb-test-cleanup.sh"],
     haltOnFailure=False, maxTime=1800, sigtermTime=30, logEnviron=False,
     lazylogfiles=True,
-    logfiles={"coverage-user"   : { "filename" : "coverage-user.tar.xz",   "follow" : False },
-              "coverage-kernel" : { "filename" : "coverage-kernel.tar.xz", "follow" : False }},
+    logfiles={"coverage-user"   : "coverage-user.tar.xz"   },
+              "coverage-kernel" : "coverage-kernel.tar.xz" }},
     decodeRC={0 : SUCCESS, 1 : FAILURE, 2 : WARNINGS, 3 : SKIPPED },
     description=["analysis"], descriptionDone=["analysis"],
     hideStepIf=lambda results, s: results==SKIPPED))
```
would it be preferable to simply state the filename? looking at the linked documentation, the default for "follow" is "False".. so I think the above simplification would be equivalent.. but maybe being explicit and not relying on defaults is better?

dinatale2:
@prakashsurya My vote is to stay consistent with the rest of `master.cfg`.

behlendorf:
If we're going to start keeping larger logs on the server we should consider setting a [log horizon](https://docs.buildbot.net/0.8.7p1/full.html#cfg-logHorizon) to age out old logs.  Keeping say the last 200 or so builds would likely be fine.

dinatale2:
Merged! Thanks @prakashsurya!

PR #113, opened 2017-09-30
codecov: fix missing redirection

prakashsurya:
The prior commit to export the HTML coverage reports is broken due to it
missing the required redirection operator. Here's what the error message
would look like:

```
+ xz -9e /var/lib/buildbot/slaves/zfs/Ubuntu_17_04_x86_64_Coverage__TEST_/build/tests/coverage-user.tar.xz
+ tar -cf - coverage-user
xz: /var/lib/buildbot/slaves/zfs/Ubuntu_17_04_x86_64_Coverage__TEST_/build/tests/coverage-user.tar.xz: No such file or directory
```

This patch attempts to fix the issue.


dinatale2:
Merged. Thanks!

PR #118, opened 2017-11-06
Retire the SIMD performance builder

dinatale2:
We have decided to retire this builder due to high cost.


dinatale2:
@behlendorf Updated the commit message to have the clarification and removed `ZFSEC2VectorTestSlave`.

PR #120, opened 2017-11-06
Allow control of zfs-tests iteration count

dinatale2:
Recent changes to zfs-tests allow multiple zfs-test iterations
to be executed. Make this accessible to developers in the
buildbot.


scripts/bb-test-zfstests.sh
```diff
@@ -38,10 +38,12 @@ set -x
 DEFAULT_ZFSTESTS_RUNFILE=${DEFAULT_ZFSTESTS_RUNFILE:-"linux.run"}
 DEFAULT_ZFSTESTS_DISKS=${DEFAULT_ZFSTESTS_DISKS:-""}
 DEFAULT_ZFSTESTS_TAGS=${DEFAULT_ZFSTESTS_TAGS:-"functional"}
+DEFAULT_ZFSTESTS_ITERS=${DEFAULT_ZFSTESTS_ITERS:-"1"}
```
behlendorf: Unlike the other `DEFAULT_` entries I don't think there's a good reason for us to allow individual builders to provide different default iterations.  Let's drop this line and move the default value in to `TEST_ZFSTESTS_ITERS` like most of the other tunings.

dinatale2:
@behlendorf good point, refreshed.

PR #123, opened 2017-11-19
libasan is required for zfs#6884.

DeHackEd:
I expect this build to not work on all platforms because I don't know
exactly which package version I need yet. Each version of gcc
rolls out new ABI versions.

### DIFF
--- a/scripts/bb-dependencies.sh
+++ b/scripts/bb-dependencies.sh
@@ -45,6 +45,11 @@ CentOS*)
         zlib-devel libuuid-devel libblkid-devel libselinux-devel \
         xfsprogs-devel libattr-devel libacl-devel libudev-devel \
         device-mapper-devel openssl-devel
+
+    # Packages that are version dependent and not always available
+    if cat /etc/centos-release | grep -Fq 7.; then
+        sudo -E yum -y install libasan
+    fi
     ;;
 
 Debian*)
@@ -62,6 +67,9 @@ Debian*)
         zlib1g-dev uuid-dev libblkid-dev libselinux-dev \
         xfslibs-dev libattr1-dev libacl1-dev libudev-dev libdevmapper-dev \
         libssl-dev
+
+    # Testing support libraries
+    sudo -E apt-get --yes install libasan
     ;;
 
 Fedora*)
@@ -78,6 +86,8 @@ Fedora*)
         libuuid-devel libblkid-devel libselinux-devel \
         xfsprogs-devel libattr-devel libacl-devel libudev-devel \
         device-mapper-devel openssl-devel
+
+    sudo -E yum -y install libasan
     ;;
 
 RHEL*)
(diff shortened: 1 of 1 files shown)


DeHackEd:
Unfortunately I can't tell if I have the right versions for anything but the Fedora, redhat and centos variants. I was hoping there would be buildbots or something I could check the output of.

In particular I'd love to see it actually try a build. Any old C program (even hello-world) with `-fsanitize=address` on the command-line.

scripts/bb-dependencies.sh
```diff
@@ -45,6 +45,11 @@ CentOS*)
         zlib-devel libuuid-devel libblkid-devel libselinux-devel \
         xfsprogs-devel libattr-devel libacl-devel libudev-devel \
         device-mapper-devel openssl-devel
+
+    # Packages that are version dependent and not always available
+    if cat /etc/centos-release | grep -Fq 7.; then
```
dinatale2: Is this check sufficient? Did CentOS 6 have patched version that would create a version number similar to `6.7.123`? If so, this check would succeed when it should fail on CentOS 6.7.

scripts/bb-dependencies.sh
```diff
@@ -102,6 +112,12 @@ RHEL*)
         libuuid-devel libblkid-devel libselinux-devel \
         xfsprogs-devel libattr-devel libacl-devel libudev-devel \
         device-mapper-devel openssl-devel
+
+    # Packages that are version dependent and not always available
+    if cat /etc/redhat-release | grep -Fq 7.; then
```
dinatale2: See comment above about version numbers.

scripts/bb-dependencies.sh
DeHackEd: CentOS does do that for 7.x, but not 6.x so it works out. Here's a small selection of version strings on systems I have access to:
<pre>
CentOS release 6.9 (Final)
Red Hat Enterprise Linux Server release 6.5 (Santiago)
CentOS Linux release 7.4.1708 (Core) 
</pre>
But since I'm targeting 7.x specifically this is fine.

dinatale2: Awesome, thanks! Just wanted to be sure.

scripts/bb-dependencies.sh
```diff
@@ -135,6 +151,8 @@ Ubuntu*)
         xfslibs-dev libattr1-dev libacl1-dev libudev-dev libdevmapper-dev \
         libssl-dev
 
+    sudo -e apt-get --yes install libasan1
```
loli10K: One of the Ubuntu builders is failing the "install dependencies" step: http://build.zfsonlinux.org/builders/Ubuntu%2014.04%20i686%20%28TEST%29/builds/4822:

```
Processing triggers for libc-bin (2.19-0ubuntu6.13) ...
+ sudo -e apt-get --yes install libasan1

Error reading /root/.nanorc: Permission denied

Error reading /root/.nano_history: Permission denied

Press Enter to continue starting nano.

command timed out: 1200 seconds without output, attempting to kill
process killed by signal 9
program finished with exit code -1
```

Maybe it's just a typo, the correct sudo option should be "-E" (capital "e").

dinatale2: @loli10K I'll get a fix going. Thanks!

dinatale2: @loli10K I pushed a couple of fixes to master. One for the `sudo -E` typo and one for the Debian package name (should've been `libasan1` instead of `libasan`). Let me know if you run into other issues.

PR #124, opened 2017-11-21
Fix "binary file log matches" error

bunder2015:
grep sometimes believes that the main log output
file is a binary file, which stops processing
of the failed/killed output at the end of the
test run.  Changing grep syntax should correct
this.

eg: http://build.zfsonlinux.org/builders/Fedora%2027%20x86_64%20%28TEST%29/builds/50/steps/shell_8/logs/stdio


bunder2015:
I believe this might actually be caused by using carets (^) in various tests.

PR #132, opened 2018-02-07
Provide a summary log for zfstests step 

dinatale2:
On the zfstests step of TEST builders, put the FAIL and
KILLED test cases into fail.log to make it easier to find
failures.

### DIFF
--- a/scripts/bb-test-zfstests.sh
+++ b/scripts/bb-test-zfstests.sh
@@ -18,17 +18,30 @@ if echo "$TEST_ZFSTESTS_SKIP" | grep -Eiq "^yes$|^on$|^true$|^1$"; then
 fi
 
 CONSOLE_LOG="$PWD/console.log"
+SUMMARY_LOG="$PWD/summary.log"
+TEST_LOG="$PWD/test.log"
 KMEMLEAK_LOG="$PWD/kmemleak.log"
 KMEMLEAK_FILE="/sys/kernel/debug/kmemleak"
 RESULT=0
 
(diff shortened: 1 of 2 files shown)


behlendorf:
And while I'm wishing for things what would be really nice is if we could extract the relevant sections from the full log.

scripts/bb-test-zfstests.sh
```diff
@@ -18,17 +18,30 @@ if echo "$TEST_ZFSTESTS_SKIP" | grep -Eiq "^yes$|^on$|^true$|^1$"; then
...
     sudo -E $ZFS_SH -u
     dmesg >$CONSOLE_LOG
+
+    if [ -f "$TEST_LOG" ]; then
+        grep -A 20 "Results Summary" "$TEST_LOG" > $SUMMARY_LOG
+        echo "" >> $SUMMARY_LOG
+        grep -a "\[KILLED\]" log >> $SUMMARY_LOG
+        grep -a "\[FAIL\]" log >> $SUMMARY_LOG
+        echo "" >> $SUMMARY_LOG
+        awk '/\[FAIL\]|\[KILLED\]/{ show=1; print; next; } ' \
```
behlendorf: Looks like the `awk` needs a little tweaking.

`awk: fatal: cannot open file `/\[SKIP\]|\[PASS\]/{ show=0; } show' for reading (No such file or directory)`

http://build-dev.zfsonlinux.org/builders/Fedora%2027%20x86_64%20%28TEST%29/builds/3

dinatale2:
@behlendorf I think I have the issues resolved.

PR #135, opened 2018-02-25
Add pyzfs runtime dependencies

loli10K:
The Foreign Function Interface for Python (python-cffi) is needed to load pyzfs (and thus to run Python unittest in the ZFS Test Suite)

The package name is "python-cffi" on every distro i've tested (Debian8, CentOS6/7, Fedora26 and Ubuntu17). The other needed Python module is "unittest", but that seems to be pulled in by the main python package.

Small question: since "python-cffi" is not a "development" packages per se do we install it only if `test "$BB_MODE" = "TEST";`?


dinatale2:
@loli10K I'm not sure it's important to add that condition. @behlendorf thoughts?

behlendorf:
Let's keep it simple as you've done.  LGTM.

PR #137, opened 2018-03-01
Fix tables' headers in html pages

gmelikov:
It's easier to work with large tables when you see it's header.

Use datatables plugin for this (scroll page to see it in action)
https://datatables.net/extensions/fixedheader/examples/options/header_footer.html

Signed-off-by: George Melikov <mail@gmelikov.ru>


gmelikov:
@dinatale2 @behlendorf I'll be glad if you could review this small change. Unfortunately I can't set you as suggested reviewers due to project rights.

PR #138, opened 2018-03-05
Further extend known-issues.sh script

behlendorf:
* Fix GitHub API query to match all issues with the "Test Suite" label.

* Added "State" column which shows the GitHub issue state ("open", or "closed").  Closed issues continue to appear on the list until they haven't been observed in the last -m days.

* Add optional -e exceptions list which behaves similarly to the OpenZFS exceptions list.  Use '-' to skip a known issue, '!' to provide a custom states, or specify a alternate issue number.

Example: http://build.zfsonlinux.org/known-issues.html


gmelikov:
I can add this color change in next PR, if you agree with this thought.

dinatale2:
@gmelikov I wouldn't mind seeing a sample of the page with the color based on state.

behlendorf:
@gmelikov please do!

PR #143, opened 2018-03-22
Add PTS performance builder

behlendorf:
* Add a bb-test-pts which runs several of the PTS "Disk" tests
  for four different zpool configurations on a dedicated
  performance builder.

* Initial version of test profiles to run for ZFS described by
  the pts-test-profiles.tar.gz tarball.  The exact performance
  values here aren't as important as our relative delta over time.


scripts/bb-test-cleanup.sh
```diff
@@ -177,6 +177,9 @@ export CODE_COVERAGE_BRANCH_COVERAGE=${TEST_CODE_COVERAGE_BRANCH_COVERAGE:-0}
 set -x
 cd "${ZFS_BUILD}"
 
+# Unload modules
+sudo -E $ZFS_SH -vu
```
dinatale2: Is this necessary anymore? The PERF builder doesn't run cleanup anymore.

master/master.cfg
```diff
@@ -1530,6 +1540,15 @@ c['schedulers'].append(Nightly(
     onlyIfChanged=False,
     dayOfWeek=6, hour=19, minute=0))
 
+# This scheduler is to trigger nightly performance tests.
+#c['schedulers'].append(Nightly(
+#    name="nightly-pts-scheduler",
```
dinatale2: Was this intentionally commented out?

behlendorf: Nope, fill fix.

scripts/bb-test-cleanup.sh
behlendorf: Right, I'll drop it.

behlendorf:
@dinatale2 I'm not sure.  That's something we'll want to investigate and may need a follow up PR to address.

dinatale2:
@behlendorf Sounds good.

PR #147, opened 2018-04-01
Handle when apt-get upgrade takes a long time

dinatale2:
On the Ubuntu builders, bb-dependencies can fail if an
apt-get upgrade runs too long. Introduce a busy loop
in bb-dependencies to prevent it from failing.


dinatale2:
@behlendorf Something like this seem like a reasonable solution?

PR #148, opened 2018-04-01
Remove tput usage in bb-dependencies

dinatale2:
`TERM` is apparently set to `Unknown` during build step execution. Simplify the `apt-get-install` to not use `tput`.

### DIFF
--- a/scripts/bb-dependencies.sh
+++ b/scripts/bb-dependencies.sh
@@ -9,23 +9,12 @@ else
    exit 0
 fi
 
-# borrowed from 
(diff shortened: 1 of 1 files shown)


scripts/bb-dependencies.sh
```diff
@@ -13,19 +13,9 @@ fi
 # https://askubuntu.com/questions/132059/how-to-make-a-package-manager-wait-if-another-instance-of-apt-is-running
```
behlendorf: I'd drop the reference here since you've simplied this down to a trival wait loop.

PR #152, opened 2018-04-10
Use apt-get install return code in bb-dependencies

dinatale2:
Use apt-get install return code to determine if buildbot
should retry package installation. Error code 11 indicates
that a lock file couldn't be obtained to install packages,
retry in those cases.

### DIFF
--- a/scripts/bb-dependencies.sh
+++ b/scripts/bb-dependencies.sh
@@ -12,12 +12,15 @@ fi
 # a function to wait for an apt-get upgrade to finish
 function apt-get-install
 {
-    echo "Waiting for other software managers to finish..."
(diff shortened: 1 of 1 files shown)


scripts/bb-dependencies.sh
```diff
@@ -12,12 +12,15 @@ fi
 # a function to wait for an apt-get upgrade to finish
 function apt-get-install
 {
-    echo "Waiting for other software managers to finish..."
-    while fuser /var/lib/dpkg/lock >/dev/null 2>&1 ; do
+    while : ; do
```
behlendorf: Let's change this to `while true; do` which is more conventional.

PR #153, opened 2018-04-12
Add support for VPC and block device configuration for build slaves

Conan-Kudo:
For hybrid cloud deployments where part of the resources are private or partitioned, it is necessary for the build slaves to be able to be configured to be instantiated with a VPC configured.

In addition, for using official cloud images from CentOS and Fedora (and potentially any other distribution's official images), buildbot needs to provision the build slave in EC2 with the ephemeral virtual disks rather than having them manually configured by cloning an AMI in the AWS console.

Signed-off-by: Neal Gompa <ngompa@datto.com>


dinatale2:
@Conan-Kudo Thanks for the PR! I’m going test it out next week. I’ll post questions if I have any.

Conan-Kudo:
@dinatale2 Here's a package built for Fedora 27+ (including Rawhide) and Mageia 6 and Cauldron for buildbot so you can test this easily: https://copr.fedorainfracloud.org/coprs/ngompa/buildbot-zfs-eight/

```bash
$ sudo dnf copr enable ngompa/buildbot-zfs-eight
$ sudo dnf install buildbot-master
# Do setup stuff...
# ???
# Profit!
```

PR #162, opened 2018-05-30
Conditionally checkout and build the spl

behlendorf:
Update the CustomGitHubEventHandler to conditionally checkout and
build the SPL when required.  This is currently possible using
the "buildspl" property, but that propery is set by the builder
and cannot by overidden by a change.

Therefore, the "buildspl" property has been switched to a "Change"
properly so it can be set correctly in CustomGitHubEventHandler.
The "buildspl" properly will be set to "yes" when building any
commit on the zfs-0.7-release branch.  Otherwise, it will be set
to "no" and the spl checkout and build will be skipped.

As part of this change the CustomGitHubEventHandler() was moved
in to its own file to simplify the master.cfg.  Some additional
minor cleanup was also included.

### DIFF
--- a/master/github.py
+++ b/master/github.py
@@ -1,0 +1,266 @@
+# -*- python -*-
+# ex: set syntax=python:
+
+import logging
+import urllib2
+import json
+import string
+import re
+
+from password import *
+from buildbot.status.web.hooks.github import GitHubEventHandler
+from dateutil.parser import parse as dateparse
+from twisted.python import log
+
+def query_url(url, token=None):
+    log.msg("Making request to '%s'" % url)
+    request = urllib2.Request(url)
+    if token:
+        request.add_header("Authorization", "token %s" % token)
+    response = urllib2.urlopen(request)
+
+    return json.loads(response.read())
+
+#
+# Custom class to determine how to handle incoming Github changes.
+#
+class CustomGitHubEventHandler(GitHubEventHandler):
+    valid_props = [
+        ('^Build[-\s]linux:\s*(yes|no)\s*$', 'override-buildlinux'),
+        ('^Build[-\s]lustre:\s*(yes|no)\s*$', 'override-buildlustre'),
+        ('^Build[-\s]spl:\s*(yes|no)\s*$', 'override-buildspl'),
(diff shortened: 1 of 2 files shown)


master/github.py
```diff
@@ -0,0 +1,255 @@
...
+                    props[prop_name] = json.dumps(m.group(1).lower())
+
+            # Extract if the commit message has property overrides
+            category = self.parse_comments(comments,
+                "style,build,test,perf,coverage,unstable")
+
+            # Releases prior to 0.8.0 required an external spl build.
+            match = re.match(r".*-0.[0-7]-release", branch)
+            if not match:
+                props['buildspl'] = json.dumps("no")
```
dinatale2: I think `json.dumps` is causing quotes around the various properties. Was that intentional?

master/master.cfg
```diff
@@ -66,15 +70,21 @@ from buildbot.status.results import SKIPPED
...
+        m = re.match(r".*yes.*", props[override_name])
+        if m is not None:
             return True
         else:
             return False
 
-    if props.hasProperty(name) and props[name] == "yes":
-        return True
+    if props.hasProperty(name):
+        m = re.match(r".*yes.*", props[name])
```
dinatale2: Was the regex match introduced because some properties are now surrounded by quotes?

master/github.py
behlendorf: Yes, it's intentional but undesirable.  The change properties are expected to be stored as json in the database.  When the change is scheduled the property value is assigned with `json.loads` which seems to preserve the quotes.  From what I can tell this seems to be a known python 2.7 issue which was fixed in newer versions of python.

I wasn't able to find a way to prevent this, so I figured we could live with it for now.  But if you know how that'd be great!

master/master.cfg
behlendorf: Yes, exactly.

master/github.py
dinatale2: Bah, that's unfortunate. If we wanted to fix it, we'd have to introduce another buildbot patch to fix it around the `json.loads` call is my guess.

PR #176, opened 2020-01-04
Add python2 and python3 packages for FreeBSD

freqlabs:
The FreeBSD AMIs have python36 installed already, but that only gives
us /usr/local/bin/python3.6, while we also need the python3 link.

We need python2 as well in some places.

Signed-off-by: Ryan Moeller <ryan@ixsystems.com>


freqlabs:
I have a hunch this is the problem, since the configure output doesn't seem to be finding python3 or python2 either.

PR #177, opened 2020-01-04
Allow repo update when installing dependencies

freqlabs:
Signed-off-by: Ryan Moeller <ryan@ixsystems.com>


freqlabs:
Something else is holding a lock on the pkg db that prevents it from being updated. Closing this until I have some idea how to deal with that.

PR #178, opened 2020-01-04
FreeBSD zfs-tests.sh debugging

freqlabs:
A few temporary changes in hopes of getting to the bottom of why zfs-tests.sh isn't running on FreeBSD.
Also trying to figure out what is interfering with pkg update.

Since it's only changing scripts there is no buildbot reload necessary and we can just retry an existing job to run it.


freqlabs:
Successfully debugged :)

PR #179, opened 2020-01-06
Fix several problems with the FreeBSD CI

freqlabs:
* Wait for the firstboot `pkg install awscli` as an extra safety measure on top of `--no-repo-update`
* Update the list of packages required as test dependencies
* Fix TEST_* variables in commit messages being lost


freqlabs:
Oops hold up I missed something

freqlabs:
Ok ready

behlendorf:
Merged.

PR #181, opened 2020-02-14
Add pam-devel / libpam0g-dev dependency

behlendorf:
Add the pam development headers to the list of dependent packages
installed by the CI.  These are required to build a pam module.


felixdoerre:
I believe that's all. I installed the full list of dependencies in a "plain" debian container and zfs compiled successfully with the pam module.

PR #189, opened 2020-04-17
Add build-only bots for FreeBSD 12 and 13

freqlabs:
Signed-off-by: Ryan Moeller <ryan@iXsystems.com>

### DIFF
--- a/master/master.cfg
+++ b/master/master.cfg
@@ -799,13 +799,29 @@ debian8_ppc64_slave = [
         slave_userpass["Debian-8-ppc64-buildslave2"]),
 ]
 
+freebsd12_amd64_slave = [
+    ZFSEC2BuildSlave(
+        name="FreeBSD-stable/12-amd64-buildslave%s" % (str(i+1)),
+        get_image=get_freebsd12_image
+    ) for i in range(numarchslaves)
+]
+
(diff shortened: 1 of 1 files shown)


master/master.cfg
```diff
@@ -799,13 +799,29 @@ debian8_ppc64_slave = [
...
+    ) for i in range(numarchslaves)
+]
+
+freebsd13_amd64_slave = [
+    ZFSEC2BuildSlave(
+        name="FreeBSD-head-amd64-buildslave%s" % (str(i+1)),
+        get_image=get_freebsd13_image
+    ) for i in range(numarchslaves)
+]
+
```
behlendorf, line 815: I'm not convinced we really want to add these.  We used to have build-only bots for all the Linux distributions as well but ended up removing them because in practice they weren't that useful.  It was always easy enough to look at the failed build step to determine why it failed.

What was useful was adding bots for non-x86_64 architectures as well as for useful but non-default configurations.  For example, adding a bot which would copy the latest version of OpenZFS in to the FreeBSD kernel proper and build it, or one which uses a FreeBSD kernel with additional debugging / profiling enabled.

freqlabs:
Fine with me, I click the details when something fails anyway :)

PR #191, opened 2020-05-15
Ensure ZFS exports file is used by FreeBSD NFS server

freqlabs:
This will be needed when eventually we get the zfs_share tests working on FreeBSD.

Normally the mountd rc script check zfs_enable and adds the ZFS exports file automatically, but we don't want to set that for the test bots because it enables many "conveniences" we aren't interested in.


freqlabs:
* tab -> spaces

PR #193, opened 2020-05-25
Add pamtester

felixdoerre:
This is a compromise to test the pam_zfs_key module in a way that does not block the test environment and also does some functionality test.

See also https://github.com/openzfs/zfs/pull/9903

### DIFF
--- a/scripts/bb-dependencies.sh
+++ b/scripts/bb-dependencies.sh
@@ -45,7 +45,7 @@ Amazon*)
         zlib-devel libuuid-devel libblkid-devel libselinux-devel \
         xfsprogs-devel libattr-devel libacl-devel libudev-devel \
         device-mapper-devel openssl-devel elfutils-libelf-devel \
-        libffi-devel libaio-devel libmount-devel pam-devel \
+        libffi-devel libaio-devel libmount-devel pam-devel pamtester \
         python-devel python-setuptools python-cffi
     ;;
 
@@ -74,7 +74,7 @@ CentOS*)
     sudo -E yum -y install kernel-devel \
         zlib-devel libuuid-devel libblkid-devel libselinux-devel \
         xfsprogs-devel libattr-devel libacl-devel libudev-devel \
-        device-mapper-devel openssl-devel libffi-devel pam-devel \
+        device-mapper-devel openssl-devel libffi-devel pam-devel pamtester \
         python-devel libaio-devel python-setuptools python-cffi \
         libyaml-devel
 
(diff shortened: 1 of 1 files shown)


scripts/bb-dependencies.sh
```diff
@@ -74,7 +74,7 @@ CentOS*)
     sudo -E yum -y install kernel-devel \
         zlib-devel libuuid-devel libblkid-devel libselinux-devel \
         xfsprogs-devel libattr-devel libacl-devel libudev-devel \
-        device-mapper-devel openssl-devel libffi-devel pam-devel \
+        device-mapper-devel openssl-devel libffi-devel pam-devel pamtester \
```
behlendorf, line 77: For CentOS `pamtester` appears to be in the epel repository so you'll need to move this down to the "Testing support libraries and tools" section which enables the epel repository.    It would be nice to move this to the utilities section for all the platforms

scripts/bb-dependencies.sh
```diff
@@ -45,7 +45,7 @@ Amazon*)
         zlib-devel libuuid-devel libblkid-devel libselinux-devel \
         xfsprogs-devel libattr-devel libacl-devel libudev-devel \
         device-mapper-devel openssl-devel elfutils-libelf-devel \
-        libffi-devel libaio-devel libmount-devel pam-devel \
+        libffi-devel libaio-devel libmount-devel pam-devel pamtester \
```
behlendorf, line 48: `pamtester` doesn't appear to be packages for the Amazon AMIs.  I think it'd be OK to skip the testing on this platform, what I'd suggest is using the `log_unsupported()` function to skip these tests when `pamtester` isn't installed.

felixdoerre: Yes, `log_unsupported` is already used, here: https://github.com/openzfs/zfs/pull/9903/files#diff-3b90db4527561b0d1d7c4a77cb7ebc55R26
I think that this is correct.

scripts/bb-dependencies.sh
```diff
@@ -211,7 +212,7 @@ RHEL*)
         libuuid-devel libblkid-devel libselinux-devel \
         xfsprogs-devel libattr-devel libacl-devel libudev-devel \
         device-mapper-devel openssl-devel libffi-devel libaio-devel \
-        pam-devel python-devel python-setuptools python-cffi
+        pam-devel pamtester python-devel python-setuptools python-cffi
```
behlendorf, line 215: For RHEL you'll need to move this in to ` # Testing support libraries and tools` as well.

PR #200, opened 2020-07-31
Update buildbot to 2.8.x

tonyhutter:
This patch does a number of things to update zfs-buildbot from buildbot 0.8.x to 2.8.x.

- Remove old patchsets
- Use "worker" terminology to match new buildbot APIs.
- Make a bunch of previously hard-coded global variable optionally set-able in password.py (for testing only).
- Correctly name the coverage tester Ubuntu 17 instead of 18
- Add nginx config to README.md and Ubuntu 20.04 setup instructions for buildbot.
- Fix get_freebsd_image(): `'NoneType' object has no attribute 'get_image'`
- Base64 encode user_data to get around error
- Update block_device_map
- Integrate isIdle() patch into zfs-buildbot.  We no longer have an idle worker state, so just check to see if a builder is available.
- Get rid of: "logHorizon/buildHorizon is deprecated" warnings
- Get rid of special chars in worker names (like '.').  This gets
  around an error: `"worker name '<worker name>' is not an identifier`
- Fix auth/authz/www stuff
- Fix `'nextWorker now takes a 3rd argument'`
- Fix error while parsing config file: cannot import name 'html'
- mergeRequests -> collapseRequests
- Make boot_device_map a list to match new API
- Fix `"builtins.TypeError: __init__() takes at least 3 arguments"`
- buildbot.status.results -> buildbot.process.results
- Misc other updates


tonyhutter:
Note, we'll need https://github.com/buildbot/buildbot/pull/5401 merged into upstream buildbot before we can pull in this PR.

master/master.cfg
```diff
@@ -747,206 +770,204 @@ def freebsd_ami(abi, version):
...
 
-freebsd13_amd64_testslave = [
-    ZFSEC2TestSlave(
-        name="FreeBSD-head-amd64-testslave%s" % (str(i+1)),
-        get_image=get_freebsd13_image
-    ) for i in range(0, numtestslaves)
+freebsd13_amd64_testworker = [
+    ZFSEC2TestWorker(
+        name="FreeBSD-head-amd64-testworker%s" % (str(i+1)),
+        ami=get_freebsd13_image()
```
freqlabs, line 903: We'll want to come up with a similar patch so this can be passed as a function again. The AMI needs to be determined when a worker is spawned rather than when the buildbot config is loaded.

FreeBSD is trying to get Amazon to provide us with an SSM parameter store, which would let us set a fixed parameter path as the AMI here and avoid this hack.

tonyhutter:
Closing this PR in favor of a substantially revamped and updated PR that upgrades the code to buildbot 3.2.0 (#233)

PR #199, opened 2020-07-24
Add some FreeBSD arm64 arch builders

ghost:
The AMI registry backend has been updated to watch for arm64 AMIs now, but there won't be anything there until the next AMIs are published Thursday next week. Leaving this as a draft until then.

PR #203, opened 2020-10-01
add debian10 arm64 slaves

jbergler:
I've noticed that there has been a small handful of issues with GPL
violations on arm64 over the past year or so. It seems like this is
something we can catch by leveraging the AWS graviton arm64 instances
for build slaves.

### DIFF
--- a/master/buildslaves.py
+++ b/master/buildslaves.py
@@ -217,9 +217,14 @@ class ZFSEC2StyleSlave(ZFSEC2Slave):
 
 # Create an HVM EC2 large latent build slave
 class ZFSEC2BuildSlave(ZFSEC2Slave):
-    def __init__(self, name, **kwargs):
+    def __init__(self, name, arch="amd64", **kwargs):
+        instance_types = {
+            "amd64": "c5d.large",
+            "arm64": "a1.xlarge"
+        }
+        assert arch in instance_types
         ZFSEC2Slave.__init__(self, name, mode="BUILD",
-            instance_type="c5d.large", max_spot_price=0.10, placement='b',
+            instance_type=instance_types.get(arch), max_spot_price=0.10, placement='b',
             spot_instance=True, **kwargs)
 
 # Create an HVM EC2 latent test slave
--- a/master/buildslaves.py
+++ b/master/buildslaves.py
@@ -220,7 +220,7 @@ class ZFSEC2BuildSlave(ZFSEC2Slave):
     def __init__(self, name, arch="amd64", **kwargs):
         instance_types = {
             "amd64": "c5d.large",
-            "arm64": "a1.xlarge"
+            "arm64": "c6g.large"
         }
         assert arch in instance_types
         ZFSEC2Slave.__init__(self, name, mode="BUILD",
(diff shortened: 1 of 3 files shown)


jbergler:
I'll call out right at the get go that I have no idea how to test this change, since I think a lot of the magic for actually running the buildbot setup requires access to some ec2 accounts and more understanding than I was able to piece together from the docs.

I'm happy to spend more time iterating on this to get it right if someone can provide me the guidance on what I can do to be more confident about the change.

jbergler:
Some examples of the types of issues I'm hoping this will help catch:
- https://github.com/openzfs/zfs/issues/10985
- https://github.com/openzfs/zfs/issues/8545

master/buildslaves.py
```diff
@@ -217,9 +217,14 @@ def __init__(self, name, **kwargs):
 
 # Create an HVM EC2 large latent build slave
 class ZFSEC2BuildSlave(ZFSEC2Slave):
-    def __init__(self, name, **kwargs):
+    def __init__(self, name, arch="amd64", **kwargs):
+        instance_types = {
+            "amd64": "c5d.large",
+            "arm64": "a1.xlarge"
```
jbergler, line 223: These instances are only available in a few regions for now (us-west-2 seems the closest to everything else)
How should this be handled?

jbergler: fixed in f197c4e

jbergler: I lied, c6g is available in us-west-1 but it lacks ephemeral storage.
c6gd has nvme storage available but is [only available](https://aws.amazon.com/about-aws/whats-new/2020/07/announcing-new-amazon-ec2-instances-powered-aws-graviton2-processors/) in us-east-* and us-west-2

behlendorf: The way things are now we'll need it to be available in us-west-1.  But if we're only doing compilation testing we don't need anything fancy.

behlendorf:
I completely agree we should start using the graviton arm64 instances to catch this kind of thing.  In fact there's a similar change for FreeBSD #199 pending.  Unfortunately testing this kind of change is a bit problematic.  I'm swamped right now, but I can probably test it out in a few weeks.

master/buildslaves.py
jbergler: Ok, I switched to c6g which is available in us-west-1 but lacks NVMe and updated the bootstrap script to handle the case where those were missing.

jbergler:
> I completely agree we should start using the graviton arm64 instances to catch this kind of thing. In fact there's a similar change for FreeBSD #199 pending. Unfortunately testing this kind of change is a bit problematic. I'm swamped right now, but I can probably test it out in a few weeks.

No worries, I pushed some more changes from testing the bootstrap scripts manually.
I get to the point where buildbot starts but fails to connect because it doesn't have valid credentials, so hopefully that bodes well for when you have some time.

jbergler:
Hey @behlendorf, no rush, just checking in here to see if you've had a chance to look at this at all / if there was anything else I can do?

behlendorf:
@jbergler the change looks reasonable but I haven't yet had a chance to give it a test.  Hhopefully in the next couple of days, sorry about the delay.

behlendorf:
Merged!  This change worked out nicely with just a few minor tweaks.  Thanks for putting this together, I've gone ahead and rolled out the new [Debian 10 arm64 builder](http://build.zfsonlinux.org/builders/Debian%2010%20arm64%20%28BUILD%29).

PR #212, opened 2021-01-22
FreeBSD: Update for 13.0-ALPHA2 branch rename

freqlabs:


behlendorf:
@freqlabs is this still needed after your other recent change?

freqlabs:
@behlendorf Yes, this change switches us to the latest 13.0 branch name for AMIs. The other change ensures we use the correct build flags on the new AMIs.

PR #217, opened 2021-03-04
FreeBSD: Use 13.0-BETA4 release images for now

freqlabs:
stable/13 snapshots do not exist at the moment, and main (14.0-CURRENT)
AMIs aren't available yet either, so we must use release images/sources
for 13 for a little while.


freqlabs:
RC1 builds are starting in a few hours but we're having problems now so I took what's available now, which is BETA4.

PR #224, opened 2021-04-28
Install libcurl-dev on non-FreeBSD buildds

nabijaczleweli:
Required for ZTS since https://github.com/openzfs/zfs/pull/11956


behlendorf:
I assume the `libfetch` devel bits are installed by default on FreeBSD which is why we don't need to install it?

behlendorf (approved):

nabijaczleweli:
All of libfetch is part of the FreeBSD "base" set, yeah.

Issue #227, opened 2021-05-16
I don't think bb-build-linux.sh actually builds ZFS in


rincebrain:
I had a long bug report written, but deleted it, so here's the short one.

I was wondering how openzfs/zfs#12056 could happen, and then I tried to reproduce it and found another incompatibility which I'll be filing shortly, so I was wondering how that could happen and checked output from the kernel.org builtin bot, and noticed that none of the CC/AS/LD/etc steps mentioned zfs, so I ran bb-build-linux.sh locally, and the resulting .config after it completed said "CONFIG_ZFS is not set" and did not contain ZFS.

I'll be opening a PR once I debug this, I just thought you probably should know.

rincebrain:
Nevermind, I wasn't passing enough flags to ./configure.

rincebrain:
Whoops, wrong bug...

PR #228, opened 2021-05-16
Actually enable CONFIG_ZFS in builds.

rincebrain:
It turns out that "make proper" drops the 'CONFIG_ZFS=y' line from
.config. I couldn't readily find any documentation about why that
would be so, so I just added it later.

Tested by: Just iterating on it on my testbed until it actually worked.

Closes: #227

Signed-off-by: Rich Ercolani <rincebrain@gmail.com>


rincebrain:
As an example illustrating the problem this is intended to solve, openzfs/zfs#12110 broke every test and buildbot...except this one.

behlendorf (approved):
Interesting, I'm fairly certain `CONFIG_ZFS` used to be preserved.  Perhaps something changed in the kernel build system and we failed to notice.  Either way, thanks for noticing and fix this.  LGTM.

PR #230, opened 2021-05-20
Missed a CentOS 8 package requirement.

rincebrain:
I didn't test that EPEL didn't drop the -packaging package between
7 and 8, and deliberately didn't try to install both packaging and
distlib, which means I broke the CentOS 8 bots when openzfs/zfs#12073 goes in. Drat.

Install it from its new home instead.

(This one _has_ been tested on CentOS 8.)


behlendorf:
That's unfortunate.  Having additional dependencies on repositories not enabled by default is going to make building your own packages a bit more awkward.

rincebrain:
The alternative is telling people to install distlib by default on EL8 and packaging on EL7, which is also perfectly feasible, I just was trying to pick one and stick with it where possible.

Issue #236, opened 2021-10-28
Buildbot could use a distro version update


rincebrain:
(I'm happy to twiddle this in my still-for-a-bit-copious free time if anyone likes...)

Currently, we build against:
* Centos 7 {x86_64} (build+test)
* Centos 8 {x86_64} (build+test)
* Centos 8 Stream {x86_64} (build+test)
* Debian 8 {armel, ppc64} (build)
* Debian 10 {arm64,x86_64} (build)
* Fedora 33 {x86_64} (build+test)
* FBSD {12,13,14}-CURRENT snapshots {x86_64} (build+test)
* Kernel.org git tip builtin {x86_64} (build)
* Ubuntu {16.04} {i386} (build)
* Ubuntu {18.04, 20.04} {x86_64} (build+test)

I would suggest at a minimum adding:
* ~Some CentOS 8 alternative that will keep getting updates after 2021, be it Rocky or Alma or w/e {x86_64} (build+test)~ Whoops, missed #234. Nice.
* Debian 11 {arm64,x86_64} (build+test)
* Fedora 35 {x86_64} (build+test)
* FreeBSD something {not little-endian} (build+test)

It might also be useful to add a couple weirder things like:
* Some Linux (kernel and OpenZFS built with Clang) (build and test)
* Some Linux (kernel + OpenZFS built with KASAN) (build and test)

Though unless you could source somewhere for regularly built Clang kernels, that could be a significant burden, and building the whole kernel and then running the test suite each time could suck...maybe a bot that runs once a day/week against master, or doesn't block merges (e.g. best effort) would be better for one or both of those?

I'd also like to know whether there's a reason not to just make at least the non-x86_64 buildbots testbots too? Does it just take far too long?

I don't know when people want to drop support for older things, so I don't know when it would make sense to drop the Debian 8 trees. (It might also make sense to try using ELTS for armel if we want to see what some users might likely be running and not just "old kernel", though it's not available for ppc64, and that should probably involve a recurring payment to Freexian if using it in a non-personal setting...)

[...]

PR #237, opened 2021-10-28
A pass at bumping the buildbot versioning

rincebrain:
Some poking toward #236's goals in life.

Bumped the AMI versions used for Debian 10 and Ubuntu 18.04/20.04,
added Debian 11, bumped Fedora 33 to 34.

No idea whether there's any way to test this short of "I hope it doesn't break the world", so...here. I tried copying what needed to be done from prior additions, but...


master/master.cfg
```diff
@@ -551,6 +560,7 @@ arch_slaves = \
     ubuntu16_i386_slave + \
     debian8_arm_slave + \
     debian10_arm64_slave + \
+    debian11_arm64_slave + \
```
behlendorf, line 563: Previously we just kept one builder per architecture to help catch anything major for that architecture.  But since these builds are quick I think we can keep both in this case.  It'll be nice to verify both kernels.

master/master.cfg
```diff
@@ -480,15 +480,17 @@ centos8_ami = "ami-0473bc9e4bfe85b1b"
 centosstream8_ami = "ami-0f377b303df4963ab" # CentOS Stream 8 1-06-2021
 
 # Provided by Debian: https://wiki.debian.org/Cloud/AmazonEC2Image
-debian10_amd64_ami = "ami-0ed1af421f2a3cf40" # Debian 10 9-9-2019
-debian10_arm64_ami = "ami-0a61b13f06119b46c" # Debian 10 9-28-2020
+debian10_amd64_ami = "ami-024fe42989cf9e876" # Debian 10 10-11-2021
+debian10_arm64_ami = "ami-0a0ab0f23e5ea98f9" # Debian 10 10-11-2021
+debian11_amd64_ami = "ami-09b4378b1d3387f81" # Debian 11 10-11-2021
+debian11_arm64_ami = "ami-041b92da75e2787a9" # Debian 11 10-11-2021
```
behlendorf, line 486: Unfortunately, I ran in to some issues trying to get the `buildslave` running on Debian 11 when installed from pip.  The 0.8 release requires an old version of python and I wasn't having any luck getting it to run using `update-alternatives`.  This will require a little more investigation.

behlendorf (commented):
Thanks, this was overdue.  I'll give a test when I get a chance.

master/master.cfg
rincebrain: That could be rough - Debian 11 dropped basically every Python 2 package except the actual interpreter, so you'd have to go slurp everything from Pip or something if you need it...

behlendorf: Fun.  Well I suppose all the more reason for us finally to roll buildbot forward to a modern version.  Then why don't we put this on the back burner for now.

behlendorf:
Closing.  We've retired the buildbot and migrated to GitHub actions for the CI.

PR #245, opened 2021-12-13
Update to Fedora 35

tonyhutter:
- Search and replace Fedora 33 -> Fedora 35 and use new AMI
- Use official packages from kojipkgs.fedoraproject.org


behlendorf:
Merged as b3a95674636ec0abe32caf8f97cdff75f67536cf

PR #246, opened 2022-01-20
FreeBSD: Tighten python dependencies regexes

freqlabs:
I noticed py38-jupyter-packaging being installed in test logs, which is
something we do not need.

Make the regex only match digits in the version an not anything in the
rest of the package name.


behlendorf (approved):
Looks good to me.

PR #247, opened 2022-01-28
FreeBSD: Use public SSM parameters for latest AMIs

freqlabs:
FreeBSD now registers the latest AMI IDs as public parameters in the AWS
SSM parameter store.  EC2 APIs accept SSM parameters in place of an AMI
and will resolve the AMI from the parameter value.  Use this to replace
the current hacked in method of determining the latest FreeBSD AMIs at
run time.

More info on the SSM parameters:
https://www.daemonology.net/blog/2021-08-31-FreeBSD-AMI-SSM-Public-Parameters.html


freqlabs:
NB: I don't have an environment set up for testing this.

behlendorf:
When I get a moment I'll see about manually testing this and will include it if it seems to work as expected.

freqlabs:
As I recall, this does not actually work because unlike the aws cli tools, the python library does not resolve SSM parameters. I had thought this happened on the AWS end, but apparently not.

This will have to be revisited to resolve the SSM parameter at runtime similar to how the current patch resolves the AMIs from JSON blobs in an S3 bucket.

PR #250, opened 2022-03-16
FreeBSD: Disable firstboot freebsd-update

freqlabs:
FreeBSD release AMIs perform a freebsd-update and reboot if necessary by
default on first boot.  We can't deal with the reboot cleanly, so opt
out of it.


gmelikov (approved):

behlendorf:
Merged as a4431358c85a1995bdccd8a4f348eba7554007af.

PR #253, opened 2022-04-11
bb-build: export MAKE

nabijaczleweli:
We already require gmake, but since https://github.com/openzfs/zfs/pull/13316 it's also required for proper bootstrapping, for which autohell defaults to just `make`; tested on FreeBSD 13-STABLE

### DIFF
--- a/scripts/bb-build-linux.sh
+++ b/scripts/bb-build-linux.sh
@@ -32,7 +32,7 @@ make prepare >>$MAKE_LOG 2>&1 || exit 1
 
 # Configure ZFS and add it to the kernel tree.
 cd $ZFS_DIR
-sh ./autogen.sh >>$MAKE_LOG 2>&1 || exit 1
+./autogen.sh >>$MAKE_LOG 2>&1 || exit 1
 ./configure --enable-linux-builtin --with-linux=$LINUX_DIR \
     --with-linux-obj=$LINUX_DIR >>$MAKE_LOG 2>&1 || exit 1
 ./copy-builtin $LINUX_DIR >>$MAKE_LOG 2>&1 || exit 1
--- a/scripts/bb-build.sh
+++ b/scripts/bb-build.sh
@@ -7,7 +7,7 @@ fi
 case "$BB_NAME" in
 FreeBSD*)
 	MAKE="gmake WITH_DEBUG=true"
-	if sysctl -n kern.conftxt | fgrep -qx $'options\tINVARIANTS'; then
+	if sysctl -n kern.conftxt | grep -Fqx $'options\tINVARIANTS'; then
 		MAKE="$MAKE WITH_INVARIANTS=true"
 	fi
 	NCPU=$(sysctl -n hw.ncpu)
@@ -44,7 +44,7 @@ fi
 
 set -x
 
-sh ./autogen.sh >>$CONFIG_LOG 2>&1 || exit 1
+./autogen.sh >>$CONFIG_LOG 2>&1 || exit 1
 
 case "$INSTALL_METHOD" in
 packages|kmod|pkg-kmod|dkms|dkms-kmod)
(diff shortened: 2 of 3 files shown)


scripts/bb-dependencies.sh
```diff
@@ -96,6 +96,7 @@ CentOS*)
             python-devel python-setuptools python-cffi python-packaging \
             python36 python36-devel python36-setuptools python36-cffi \
             python36-packaging
+        sudo -E yum -y install "https://vault.centos.org/centos/8/AppStream/x86_64/os/Packages/automake-1.16.1-7.el8.noarch.rpm"
```
behlendorf, line 99: Depending on a non-default automake for CentOS 7 is going to be problematic for any user building their own packages from git.  CentOS 7 is still widely deployed and supported for RedHat so it'd be preferable not to add new requirements (that's what people like about it).  Is this really required?

nabijaczleweli: Sorry, yeah, I forgot to note this. [Anything since 1.14 (Jun 2013) will do](https://github.com/openzfs/zfs/pull/13316#issuecomment-1095329829), and this is, logistically, the simplest way to get that, I think. It's *possible* to work around it, but it's gonna be awful for precisely the reason outlined in ba25a9f1c295d7799575afa39c4c9854a85ad8e3 – we'd be hardcoding the lead-up path to *every* file (e.g. in `lib/libzfs/Makefile.am` we'd be saying `lib/libzfs/libzfs_changelist.c` instead of `%D%/libzfs_changelist.c`; in `tests/zfs-tests/cmd/Makefile.am` instead of
```
scripts_zfs_tests_bin_PROGRAMS += %D%/file_check %D%/file_trunc %D%/file_write %D%/largest_file %D%/randwritecomp
%C%_file_check_SOURCES    = %D%/file/file_check.c
%C%_file_trunc_SOURCES    = %D%/file/file_trunc.c
%C%_file_write_SOURCES    = %D%/file/file_write.c
%C%_largest_file_SOURCES  = %D%/file/largest_file.c
%C%_randwritecomp_SOURCES = %D%/file/randwritecomp.c
```
we'd be saying 
```
scripts_zfs_tests_bin_PROGRAMS += tests/zfs-tests/cmd/file_check tests/zfs-tests/cmd/file_trunc tests/zfs-tests/cmd/file_write tests/zfs-tests/cmd/largest_file tests/zfs-tests/cmd/randwritecomp
tests_zfs_tests_cmd_file_check_SOURCES    = tests/zfs-tests/cmd/file/file_check.c
tests_zfs_tests_cmd_file_trunc_SOURCES    = tests/zfs-tests/cmd/file/file_trunc.c
tests_zfs_tests_cmd_file_write_SOURCES    = tests/zfs-tests/cmd/file/file_write.c
tests_zfs_tests_cmd_largest_file_SOURCES  = tests/zfs-tests/cmd/file/largest_file.c
tests_zfs_tests_cmd_randwritecomp_SOURCES = tests/zfs-tests/cmd/file/randwritecomp.c
```
which is *horrible*, bordering on unacceptable). So if you think this is an acceptable trade-off, then it'd be simplest to Do This. If not, in writing this I realised I can *probably* paper over this in autogen.sh, since the configure step doesn't use the `.am`s, so I can remove this hunk from this PR and we can revisit this if I fail?

behlendorf: If we can do something in `autogen.sh` for automake versions older 1.14 that should be acceptable.  Unfortunately, I don't think we can outright require version 1.14+ or greater just yet.

nabijaczleweli: I'm very happy to report I've made it work with CentOS 7's native automake via a paper-based heading in autogen.sh, so the only thing we need for that PR is the explicit MAKE=gmake on FreeBSD.

behlendorf: Great.  Then since you've already updated this PR to drop that commit I think we can safely go ahead and merge it.  Unless there are other changes needed?

nabijaczleweli: From what I see on the CI, I'm pretty sure this is all.

behlendorf (approved):

